PR #98, opened 2020-06-08
Add about/versioning to about section's navbar

eguiraud:


eguiraud:
@couet is this good to go?

PR #117, opened 2020-06-09
Add favicons

stwunsch:
- Replace favicon.ico in the toplevel dir with the same as for the forum
  with transparent background
- Add to the header of each page the modern way to include favicons in 3
  size (ico files are kind of deprecated, though kept due to backwards
  compatibility)

The path points to the location for the current structure and has to be adapted once we move the deployment to a setup without the `/web` prefix.


eguiraud:
I think we will always deploy into a `/web` directory, only difference will be that we put a reverse proxy in front, but it should be fine to assume we deploy in `/web`

Axel-Naumann:
The reverse proxy will make this accessible as `/`, not `/web` - not sure whether we need anything here?

Axel-Naumann:
But the png-s are a good idea either way!

eguiraud:
> The reverse proxy will make this accessible as `/`, not `/web` - not sure whether we need anything here?

access to the favicon should also be reverse-proxied. i expect all content available under `/web` to be served from `/` by the reverse proxy, and just work. if I'm wrong, i'll fix it.

stwunsch:
I would also just put this in for now, keep the issue open and fix the path for the finally deployed version.

eguiraud:
Let's merge -- if changing the URL breaks the favicons I'll fix the URLs they are available from

PR #149, opened 2020-06-10
Add 'Setting the C++ standard' to 'build_from_source.md'

eguiraud:
Also mention our STL backports in this section.

@Axel-Naumann let me know if the info here is ok, especially for what regards the backports.

### DIFF
--- a/install/build_from_source.md
+++ b/install/build_from_source.md
@@ -114,13 +114,21 @@ For ROOT <= 6.20, an older version of PyROOT (not based on Cppyy) will be built.
 
 ## Setting the C++ standard
 
-> TODO: explain how ROOT features/STL backports interact with C++ standards and how to pick a standard
+ROOT needs to be configured and built with the same C++ standard as the programs that will make use of it.
+The relevant cmake flag is [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html){:target="_blank"}.
+For example, from the command line, the standard can be selected by passing one of `-DCMAKE_CXX_STANDARD={11,14,17,...}`.
+
+#### ROOT STL backports
+
+ROOT backports certain useful C++ standard library features to make them available in C++11, for example `std::make_unique` and `std::string_view`.
+The backports can be found [here in the reference guide](https://root.cern.ch/doc/master/dir_7780993579c9aa6baf9598fd7cc29d54.html).
+The backports fallback to the actual STL implementation if the C++ standard ROOT is compiled with supports the feature.
 
 ## Enabling experimental features, aka ROOT7
 
 New and improved versions of standard ROOT components are being implemented under codename ROOT7.
 
-Turning on C++14 or higher using `-DCMAKE_CXX_STANDARD={14,17,...}` will automatically enable ROOT7.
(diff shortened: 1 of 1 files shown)


install/build_from_source.md
```diff
@@ -114,13 +114,21 @@ For ROOT <= 6.20, an older version of PyROOT (not based on Cppyy) will be built.
 
 ## Setting the C++ standard
 
-> TODO: explain how ROOT features/STL backports interact with C++ standards and how to pick a standard
+ROOT needs to be configured and built with the same C++ standard as the programs that will make use of it.
+The relevant cmake flag is [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html){:target="_blank"}.
+For example, from the command line, the standard can be selected by passing one of `-DCMAKE_CXX_STANDARD={11,14,17,...}`.
```
Axel-Naumann, line 119: Our users love copy&paste, so maybe:
```suggestion
For example, from the command line, the standard can be selected by passing one of 11, 14, 20,... such as `-DCMAKE_CXX_STANDARD=14`.
```

install/build_from_source.md
```diff
@@ -114,13 +114,21 @@ For ROOT <= 6.20, an older version of PyROOT (not based on Cppyy) will be built.
...
+
+#### ROOT STL backports
+
+ROOT backports certain useful C++ standard library features to make them available in C++11, for example `std::make_unique` and `std::string_view`.
+The backports can be found [here in the reference guide](https://root.cern.ch/doc/master/dir_7780993579c9aa6baf9598fd7cc29d54.html).
```
Axel-Naumann, line 124: AFAIK that URL is not stable :-( Or is it? Could you check for v616 vs master?

install/build_from_source.md
```diff
@@ -114,13 +114,21 @@ For ROOT <= 6.20, an older version of PyROOT (not based on Cppyy) will be built.
...
+The backports fallback to the actual STL implementation if the C++ standard ROOT is compiled with supports the feature.
```
Axel-Naumann, line 125:
```suggestion
The backports are disabled, falling back to the actual C++ standard library implementation if it provides it, depending for instance on the C++ standard ROOT is compiled with and the compiler version.
```

install/build_from_source.md
```diff
@@ -114,13 +114,21 @@ For ROOT <= 6.20, an older version of PyROOT (not based on Cppyy) will be built.
...
 
 ## Enabling experimental features, aka ROOT7
 
 New and improved versions of standard ROOT components are being implemented under codename ROOT7.
 
-Turning on C++14 or higher using `-DCMAKE_CXX_STANDARD={14,17,...}` will automatically enable ROOT7.
+Turning on C++14 (see [Setting the C++ standard](#setting-the-c-standard)) or higher using  will automatically enable ROOT7.
```
Axel-Naumann, line 131:
```suggestion
Turning on C++14 (see [Setting the C++ standard](#setting-the-c-standard)) or higher will automatically enable ROOT7.
```

install/build_from_source.md
eguiraud: It happens to work if you substitute `master` with any of `v616`,`v618`,`v618`,`v620`,`v622`.

If there is a better way to link to pages in our reference or a better list of our backports we should use that though.

PR #159, opened 2020-06-11
Make about section navbar a flat list of useful links

eguiraud:
No need to repeat the page text in the navbar.

Partially addresses #8


couet:
Well .... if you do you will oblige people to go back to read the next paragraph ... Fine with me but with the current version you can scan the paragraph just by clicking in the nav bar

eguiraud:
makes sense!

Issue #97, opened 2020-06-08
Ambiguous "Get started" and "Getting started with ROOT" pages


eguiraud:
We have both: https://root-project.github.io/web/manual/getting_started_with_root and https://root-project.github.io/web/get_started .

Maybe the former could be called "First steps with ROOT" or something like that? Or: do we need both?

Axel-Naumann:
Changes URLs, we should decide before publication. I was also surprised - I like your suggestion of renaming one.

Axel-Naumann:
Agreed to rename manual page to "First steps with ROOT"

PR #175, opened 2020-06-16
{Getting Started --> First steps} with ROOT

vepadulano:
Addresses #97


eguiraud:
Thank you @vepadulano ! (tip: if you write 'Fixes #97' instead, github picks that up and closes the issue automatically)

PR #204, opened 2020-06-17
Add HTML proofer action

stwunsch:


eguiraud:
could also use https://github.com/marketplace/actions/htmlproofer , not sure what runs faster

stwunsch:
I decided against using the action since I'm not yet sure how we'll do the report. Up to now I would just pipe the output in a file and throw a grep on it.

stwunsch:
@eguiraud I would take the action as it is into the `gh-pages` branch, for now. Actually, I would not block the deployment because of a failing HTML proofer test, though I would rather block PRs if the check fails.

stwunsch:
And then running again the HTML proofer on gh-pages, even though it should be green by construction.

stwunsch:
Also we have to fix the internal stuff first and then potentially we can enable also external link checking. But probably that breaks the action such as it did for the previous try with liche.

PR #206, opened 2020-06-18
Fixing links, part 2

stwunsch:


eguiraud:
Thanks a lot!

stwunsch:
There are still the failures on the `page2`, `page3`, ... pages. I've not yet figured out why they are there and what they are doing.

eguiraud:
it turns out that they are auto-generated by jekyll, they are not leftovers of tests as I first thought. i hope it's possible to turn that off...

stwunsch:
Down to 11 failures (in the internal links :see_no_evil: )

stwunsch:
I can veto these additional pages with a regex, like as the last solution.

eguiraud:
it's this: https://jekyllrb.com/docs/pagination/

eguiraud:
you can try playing with the `paginate` option in `_config.yml`

manual/graphs/index.md
```diff
@@ -11,7 +11,7 @@ A graph is an object made of two arrays `X` and `Y`, holding the `x`,`y` coordin
 **Un-binned data**
 
 A graph or chart is a set of categorical variables, this is un-binned data.<br>
-A histogram is used for continuous data, where the bins represent ranges of data (binned data), see → [Histograms]({{ '/manual/histograms' | relative_url }}).
+A histogram is used for continuous data, where the bins represent ranges of data (binned data), see → [Fitting histograms](/manual/fitting).
```
eguiraud, line 14: we probably don't use them consistently enough, but links generated with `{{ .. | relative_url }}` are the recommended way, see https://jekyllrb.com/docs/github-pages/#project-page-url-structure

stwunsch: Thanks, good to know! But do we care about this? `jekyll serve` does the right thing, the linkchecker understand the simpler way to put links and also the deployed website is happy.

stwunsch:
Down to 1 error! ->

```
Ran on 99 files!
- build/manual/python/bi_changes/index.html

  *  linking to internal hash #main that does not exist (line 0)

     <a href="#main" class="screen-reader-shortcut">Skip to content</a>
htmlproofer 3.15.3 | Error:  HTML-Proofer found 1 failure!
```

manual/graphs/index.md
eguiraud: i don't know what will break without the relative_url, but if they say that's the right way, why should we go out of our way to do it differently? :D

as an example, here is an account of things went wrong without relative_url: https://github.com/github/pages-gem/issues/460

stwunsch: But that's only for the baseurl `"/"`. But fair enough, let me make an issue to keep track.

stwunsch: https://github.com/root-project/web/issues/208

stwunsch:
It's greeeee! :partying_face:

Issue #210, opened 2020-06-18
Check external links with the HTML proofer


stwunsch:
I've disabled the check of external links by the HTML proofer due to massive failures in the test deployment. We should try to reenable the check of external links once the deployment is fix.

stwunsch:
Fixed by https://github.com/root-project/web/pull/212

Issue #194, opened 2020-06-17
Links broken in manual/vectors


stwunsch:
We have broken links in `/manual/vectors`, see here: https://root-project.github.io/web/manual/vectors/

The issue is that the `ref` mechanic does not allow to take the namespaces of classes into account since doxygen maps `::` to `_1_1`. Also, some links that work have hard coded links without using the `ref` mapping. So what should be the correct way to fix it?

See here for the source of the page: https://raw.githubusercontent.com/root-project/web/gh-pages/manual/vectors/index.md

stwunsch:
Fixed by https://github.com/root-project/web/pull/212

PR #212, opened 2020-06-18
enable check of extern links

stwunsch:


stwunsch:
Fixes https://github.com/root-project/web/issues/210

stwunsch:
Fixes https://github.com/root-project/web/issues/194

stwunsch:
The remaining errors are desy, which return 403 though the URL is clearly accessible without restrictions. I'll merge for now and investigate the issue separately.

PR #216, opened 2020-06-18
enable httpproofer on push

stwunsch:
Enable the HTTP proofer on push to gh-pages

### DIFF
--- a/.github/workflows/html-proofer.yml
+++ b/.github/workflows/html-proofer.yml
@@ -1,6 +1,8 @@
 name: HTML proofer
 
 on:
+  push:
+    branches: [ gh-pages ]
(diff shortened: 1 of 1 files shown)


stwunsch:
It's greeeeeen! :green_heart:

stwunsch:
It seems that this would be ready for gh-pages. However, indico.desy.de seems to block the checks if we run too often :/ But just the assumption from the 403 error code.

stwunsch:
Issues regarding indico.desy.de fixed in #224

eguiraud:
Superseded by #227  ?

stwunsch:
Yep, definitely

PR #227, opened 2020-06-18
Build and deploy PRs to subdirectories on S3

eguiraud:

PR #228, opened 2020-06-19
fix relative links of the icons

stwunsch:
The additional ticks are required to make the link work!

### DIFF
--- a/index.html
+++ b/index.html
@@ -14,7 +14,7 @@ header:
 <div style="width: 100%;">
     <center>
     <div style="width: 25%; display:inline-block;">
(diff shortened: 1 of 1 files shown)


stwunsch:
An external link is unresponsive with error code 503 :/ Probably we should enable that we fail only on 40x error codes?

stwunsch:
Screenshot from 2020-06-19 08-17-21

stwunsch:
Fixed by https://github.com/root-project/web/pull/227, the PR here can just be closed once the other one is merged. I"ll keep it open for now because the links of the icons are crucial.

eguiraud:
Fixed in #227

PR #229, opened 2020-06-19
fail html proofer only due to 4xx error codes

stwunsch:
@eguiraud What do you think? We have spurious failures of the CI, e.g., https://github.com/root-project/web/pull/228, just because another server is not responsive.

Screenshot from 2020-06-19 08-22-01


stwunsch:
The PR is most likely superseded by https://github.com/root-project/web/pull/227

eguiraud:
Done in #227

PR #263, opened 2020-06-23
Add new content to the PyROOT manual section

etejedor:
In order to have some new content when the new website is published (the section is not complete yet).


manual/python/index.md
```diff
@@ -1,70 +1,95 @@
...
 layout: single
 sidebar:
   nav: "manual"
 toc: true
 toc_sticky: true
 ---
 
-ROOT provides an interface to Python, which is enabled via a set of bindings called PyROOT.<br>
-PyROOT is enabled by default in ROOT.
+ROOT can also be used from Python thanks to its Python-C++ bindings, called **PyROOT**.  PyROOT allows to access all the ROOT functionality from Python while, at the same time, benefiting from the performance of the ROOT C++ libraries.
```
stwunsch, line 10: Probably we should point out more prominently that pyroot is not ROOT on Python but **generic** and totally automatic python bindings for C++ libraries. There's a huge difference! I think it's really hard to understand what PyROOT is capable to do, or let's say cppyy?

manual/python/index.md
```diff
@@ -1,70 +1,95 @@
...
+%}
 
-## Running Python from the ROOT/Cling interpreter
+> **Note that** the code above can also be written in a script file and executed with Python. In that case, the script will run to completion and the Python process will terminate, making the created canvas disappear. If you want to keep the Python process alive and thus be able to inspect your canvas, run the script with `python -i script_name.py`.
 
-In ROOT can run any Python command via the {% include ref class="TPython" %} class.
+## Interface
 
- _**Example**_
+Coming soon!
```
stwunsch, line 63: I would write `Content is coming soon!`, so that it's not confused with a missing feature in PyROOT.

manual/python/index.md
```diff
@@ -1,70 +1,95 @@
...
 
-{% highlight C++ %}
-root [0] TPython::Exec( "print 1 + 1" )
+## Loading user libraries & jitting
+
+Coming soon!
```
stwunsch, line 67: Same here, make clear we are just using the content on the webiste, not the actual feature.

manual/python/index.md
etejedor: Sure! I will make that clear in this intro package, thanks. I also have a section (not written yet) that will demonstrate how to dynamically load C++ libraries and call into them right away from Python.

manual/python/index.md
etejedor: Yes will do!

PR #278, opened 2020-06-24
Fix broken links

stwunsch:


eguiraud:
These links are actually correct (that's why html proofer doesn't say anything) but our webserver somehow removes the trailing slash from URLs that is required for relative URLs to work

stwunsch:
Thanks for figuring out!

PR #296, opened 2020-06-29
[PyROOT] Are more sub-sections to manual

etejedor:
This PR adds one subsection about dynamically loading C++ with PyROOT and another one for JupyROOT.

### DIFF
--- a/manual/python/index.md
+++ b/manual/python/index.md
@@ -70,7 +70,7 @@ PyROOT allows to use *any C++ library* from Python, not only ROOT. This is possi
 
 ### Jitting strings with code
 
-ROOT comes with a C++ interpreter to which we can request to process a given piece of C++ code. Sometimes, if such code is short (e.g. the definition of a small function or class) or for rapid exploration/debugging, it is enough the place the C++ code in a Python string and give it to the interpreter. The code will be just-in-time compiled (jitted) and immediately available for invocation, as shown in the example below.
+ROOT comes with a [C++ interpreter]({{ '/manual/first_steps_with_root/#using-the-interactive-c-interpreter-cling' | relative_url }}) to which we can request to process a given piece of C++ code. Sometimes, if such code is short (e.g. the definition of a small function or class) or for rapid exploration/debugging, it is enough the place the C++ code in a Python string and give it to the interpreter. The code will be just-in-time compiled (jitted) and immediately available for invocation, as shown in the example below, which calls the constructor of the C++ class `A` and function `f` from Python after defining them via the interpreter.
 
 ```python
 import ROOT
@@ -91,7 +91,7 @@ public:
 ROOT.gInterpreter.ProcessLine(cpp_code)
 
 # We find all the C++ entities in Python, right away!
-a = ROOT.A()   # this prints Hello PyROOT!
+a = ROOT.A()   # prints Hello PyROOT!
 x = ROOT.f(3)  # x = 9
 ```
 
@@ -115,7 +115,7 @@ We can execute the code below:
  ROOT.gInterpreter.ProcessLine('#include "my_header.h"')
 
  # We find all the C++ entities in Python, right away!
- a = ROOT.A()   # this prints Hello PyROOT!
+ a = ROOT.A()   # prints Hello PyROOT!
  x = ROOT.f(3)  # x = 9
 ```
 
@@ -150,7 +150,7 @@ ROOT.gInterpreter.ProcessLine('#include "my_header.h"')`
 ROOT.gSystem.Load('./my_cpp_library.so')
 
 # We find all the C++ entities in Python, right away!
-a = ROOT.A()   # this prints Hello PyROOT!
+a = ROOT.A()   # prints Hello PyROOT!
 x = ROOT.f(3)  # x = 9
 ```
 
@@ -202,7 +202,7 @@ Many ROOT tutorials are available in the form of Jupyter notebooks. For example,
 
 The ROOT graphics are also available in Jupyter, both in Python and C++. Moreover, the user can choose between two modes:
 - *Static* (default): the graphics are displayed as a static image in the output of the notebook cell.
-- *Interactive*: the graphics are shown as an interactive JavaScript display. In order to activate this mode, the `%jsroot on` line needs to be added to the cell. This is an example for a Python notebook:
+- *Interactive*: the graphics are shown as an interactive JavaScript display. In order to activate this mode, the `%jsroot on` line needs to be included in a cell. Once enabled, the mode will stay on until it is disabled with `%jsroot off` (i.e. no need to enable it in every cell). This is an example for a Python notebook:
 
 ```python
 %jsroot on
--- a/manual/python/index.md
+++ b/manual/python/index.md
@@ -66,7 +66,93 @@ caption="Example of graphics generated with PyROOT."
 
 ## Loading user libraries & jitting
 
-*Content is coming soon!*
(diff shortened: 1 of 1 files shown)


manual/python/index.md
```diff
@@ -66,9 +66,95 @@ caption="Example of graphics generated with PyROOT."
 
 ## Loading user libraries & jitting
 
-*Content is coming soon!*
+PyROOT allows to use *any C++ library* from Python, not only ROOT. This is possible thanks to the automatic and dynamic bindings between Python and C++ that PyROOT provides: without any prior generation of wrappers, at execution time, PyROOT can load and call into C++ code. The following sub-sections cover different supported scenarios for dynamic C++ code invocation.
```
stwunsch, line 69: We should point here somewhere to the cppyy docs. Or did you point to them somewhere else?

manual/python/index.md
```diff
@@ -66,9 +66,95 @@ caption="Example of graphics generated with PyROOT."
...
+
+### Jitting strings with code
+
+ROOT comes with a C++ interpreter to which we can request to process a given piece of C++ code. Sometimes, if such code is short (e.g. the definition of a small function or class) or for rapid exploration/debugging, it is enough the place the C++ code in a Python string and give it to the interpreter. The code will be just-in-time compiled (jitted) and immediately available for invocation, as shown in the example below.
```
stwunsch, line 73: Put a link to cling and/or the cling docs and/or the respective part in the manual!

manual/python/index.md
```diff
@@ -66,9 +66,95 @@ caption="Example of graphics generated with PyROOT."
...
+public:
+    A() { cout << "Hello PyROOT!" << endl; }
+};
+"""
+
+# Inject the code in the ROOT interpreter
+ROOT.gInterpreter.ProcessLine(cpp_code)
+
+# We find all the C++ entities in Python, right away!
+a = ROOT.A()   # this prints Hello PyROOT!
```
stwunsch, line 94: Probably `this calls the constructor of class A and prints Hello PyROOT!`? To make very clear what happens. I can imagine that not everyone sees right away that here a construction happens.

manual/python/index.md
```diff
@@ -66,9 +66,95 @@ caption="Example of graphics generated with PyROOT."
...
+int f(int i) { return i*i; }
 
-## IPython: running Python from C++
+class A {
+public:
+    A() { cout << "Hello PyROOT!" << endl; }
+};
+```
+
+We can execute the code below:
```
stwunsch, line 111: That's a duplication, just say: We can include a header with code `bla` and then we can run exactly the same example as above, e.g., `ROOT.f(3)`.

stwunsch: Ok, I read through all of it. You probably want the repetition to make very clear that you get the same outcome in different ways of loading the impls. So, fine by me to keep it as it is!

manual/python/index.md
```diff
@@ -87,9 +173,53 @@ root [2] TPython::Prompt()
...
 For a more complete description of the `TPython` interface, please check the reference guide for {% include ref class="TPython" %}.
 
-## JupyROOT: PyROOT for Jupyter notebooks
+## JupyROOT: (Py)ROOT for Jupyter notebooks
+
+[Jupyter notebooks](https://jupyter.org/) provide a web-based interface that can be used to do interactive computing with ROOT.
+
+### How to use it
+
+Two kernels (or language flavours) allow to run ROOT from Jupyter: *Python* (2 and 3) and *C++*. In order to use such kernels, there are mainly two options:
```
stwunsch, line 182: Is this fully correct? Is a kernel in Jupyter speech a language (like python or cpp) or a specific instance like py2, py3 or the root cpp kernel. I think the latter? In case this is true, it's not 100% technically correct to say "Two kernels ..."

manual/python/index.md
```diff
@@ -87,9 +173,53 @@ root [2] TPython::Prompt()
...
+
+- Local ROOT installation: to install ROOT locally, please follow the instructions [here]({{ '/install' | relative_url }}). In addition to ROOT, two Python packages are required, which can be installed e.g. with `pip`:
+
+```shell
+pip install jupyter
+pip install metakernel
+```
+
+Once all requirements are installed, a Jupyter server with the Python and C++ kernels can be started by running from a terminal:
```
stwunsch, line 191: How is `root --notebook` any different from `jupyter notebook`? I can also access the cpp kernel from there. I've never experienced any advantage of `root --notebook` compared to `jupyter notebook`. In case there is any difference, it might be worth to point them out here.

manual/python/index.md
```diff
@@ -87,9 +173,53 @@ root [2] TPython::Prompt()
...
+
+```shell
+root --notebook
+```
+
+- SWAN: CERN provides an online service to create and run Jupyter notebooks, called [SWAN](https://swan.cern.ch). With this option no installation is required: a browser and a working Internet connection are enough. Both the Python and C++ kernels are available in SWAN.
```
stwunsch, line 197: `internet`, lower case

manual/python/index.md
```diff
@@ -87,9 +173,53 @@ root [2] TPython::Prompt()
...
+
+Many ROOT tutorials are available in the form of Jupyter notebooks. For example, most [PyROOT tutorials](https://root.cern/doc/master/group__tutorial__pyroot.html) are listed with two buttons below to obtain their notebook version and to open them in SWAN, respectively.
+
+### JavaScript graphics
+
+The ROOT graphics are also available in Jupyter, both in Python and C++. Moreover, the user can choose between two modes:
+- *Static* (default): the graphics are displayed as a static image in the output of the notebook cell.
+- *Interactive*: the graphics are shown as an interactive JavaScript display. In order to activate this mode, the `%jsroot on` line needs to be added to the cell. This is an example for a Python notebook:
```
stwunsch, line 205: You need the magic only once per notebook, not in each cell, right? Just because there are also magics which you have to place per cell.

manual/python/index.md
```diff
@@ -87,9 +173,53 @@ root [2] TPython::Prompt()
...
+c = ROOT.TCanvas()
+f1 = ROOT.TF1("func1", "sin(x)", 0, 10)
+f1.Draw()
+c.Draw() # Necessary to make the graphics show!
+```
+
+Executing the code above in a cell will make the following interactive canvas appear as output:
+
+{% include figure_jsroot
+   file="/manual/graphics/graphics.root" object="func1" width="500px" height="350px"
```
stwunsch, line 218: Niiiice, I want to see it! :D

manual/python/index.md
etejedor: Yes, I point to the cppyy docs at the beginning of the PyROOT section.

manual/python/index.md
etejedor: Done!

manual/python/index.md
etejedor: I made a more verbose explanation of what happens in the text before the example, to not make the comment too long in the code.

manual/python/index.md
etejedor: Yes, I think defining a kernel as a language flavour is accurate:

https://jupyter.readthedocs.io/en/latest/projects/kernels.html

it is true that you see the message "starting kernel" in Jupyter, but it rather means "kernel process" in that case.

manual/python/index.md
etejedor: `root --notebook` installs the ROOT C++ kernel in the home directory of the user (if not there already) and sets the environment so that Jupyter finds it. If you run `jupyter notebook` but you have never run `root --notebook` before, you should not see the ROOT C++ kernel unless you install it manually. That is why here I talk about `root --notebook`.

manual/python/index.md
etejedor: But Internet is a proper noun, I think uppercase is better.

manual/python/index.md
etejedor: Rigtht, I should make that clear, thanks! I added a sentence that explains that.

PR #307, opened 2020-07-02
[PyROOT] Add sub-section about user interface

etejedor:


manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
-*Content is coming soon!*
+Besides being the entry point to all ROOT functionality, the `ROOT` Python module also offers an interface to configure PyROOT and to manipulate PyROOT objects at a lower level. The next subsections describe in more detail the characteristics of the PyROOT user interface.
+
+### Configuration options
+
+After importing the ROOT module, one can access the PyROOT configuration object as `ROOT.PyConfig`. Such object has a set of properties that can be modified to steer the behaviour of PyROOT. For the configuration to be taken into account, it needs to be applied right after `ROOT` is imported. For example:
+
+```python
+import ROOT
+ROOT.PyConfig.OptionName1 = SomeValue1
```
stwunsch, line 73: Put some real options here!

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+If the user would like to completely disable the rootlogon functionality, they can do so by setting `PyConfig.DisableRootLogon` to `True`. 
+
+- `IgnoreCommandLineOptions` (default `True`): if a PyROOT script is executed with some command line arguments, they will be ignored by default by ROOT, so the user is free to process them as they wish. However, by setting `PyConfig.IgnoreCommandLineOptions` to `False`, those arguments will be forwarded to ROOT for parsing, e.g. to enable the batch mode from the command line. A complete list of the arguments accepted by ROOT can be found [here]({{ '/manual/first_steps_with_root/#starting-root-with-command-line-options' | relative_url }}).
+
+- `ShutDown` (default `True`): when PyROOT is terminating, during its cleanup phase, the ROOT C++ interpreter is shut down. If PyROOT is executed as part of a longer-running application that needs the C++ interpreter, `PyConfig.Shutdown` can be set to `False` to prevent that shutdown.
```
stwunsch, line 95: Is it `ShutDown` or `Shutdown`?

stwunsch: One is wrong!

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+- `StartGUIThread` (default `True`): when not executing in [interactive mode](https://docs.python.org/3/tutorial/interpreter.html#interactive-mode), PyROOT starts a thread that periodically polls for ROOT events (e.g. GUI events) to process them. If a given PyROOT application does not need this event processing, it can prevent the creation of the thread by setting `PyConfig.StartGUIThread` to `False`.
```
stwunsch, line 97: Interactive mode or from a Jupyter notebook would be technically correct. I don't know, is it important?

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+2. Calling `gROOT.SetBatch` in the PyROOT script, right after importing `ROOT`:
+
+```python
+import ROOT
+ROOT.gROOT.SetBatch(True)
+```
+
+### Low-level manipulation of objects
+
+When instantiating a C++ class from Python via PyROOT, both a C++ object and its Python proxy object are created. Such Python object forwards any access to its internal C++ object, thus acting as a proxy. PyROOT offers some functions to inspect or manipulate Python proxies and their C++ counterparts, based on functionality provided by [cppyy](https://cppyy.readthedocs.io/en/latest/lowlevel.html). All these functions can be accessed as `ROOT.NameOfFunction` and they are listed next:
```
stwunsch, line 114: ... can be accessed with the pattern `ROOT.NameOfFunction` ...

Just to prevent that people think `ROOT.NameOfFunction` is an actual member of ROOT.

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+- `AddressOf(obj)`: when applied to a Proxy object `obj`, it returns an indexable buffer of length 1, whose only element is the address of the C++ object proxied by `obj`. The address of the buffer is the same as the address of the address of the C++ object. This function is kept for backwards compatibility with old PyROOT versions.
```
stwunsch, line 116: typo: same as the adrress of the address

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+- `addressof(obj, field, byref)`: similarly to `AddressOf`, it can be used to obtain the address of an internal C++ object from its Python proxy. However, `addressof` returns that address as an integer, not as an indexable buffer. Furthermore, `addressof` accepts two more parameters: `field` and `byref`. `field` can be used to specify the name of a field in a struct, in order to get the address of that field. If `byref` is set to true, `addressof` returns the address of the address of the C++ object. Equivalent to [cppyy's `addressof`](https://cppyy.readthedocs.io/en/latest/lowlevel.html#capsules). Example of usage:
```
stwunsch, line 118: There's something missing between here: ... the address of the C++ object. Equivalent to  ...

stwunsch: I mean to make a valid sentence out of it

stwunsch: Example of usage -> Example usage

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+94352024283040L
+>>> ROOT.addressof(mys, 'foo') # Address of "foo" field (same as address of object) 
+94352024283040L
+>>> ROOT.addressof(mys, 'bar') # Address of "bar" field
+94352024283044L
+>>> ROOT.addressof(mys, byref = True) # Address of address of mys' C++ object
+140376843834640L
+```
+
+Moreover, `addressof` can be used in conjunction with `TTree::Branch` from Python as shown [here](https://github.com/root-project/root/blob/d1d035e17b9b8dd97bcd146dd6e0c84a0d1aa4a1/bindings/pyroot/pythonizations/test/ttree_branch.py#L134).
```
stwunsch, line 135: This link will point in some near or far future to a wrong line. Probably link only the file and describe in words where to look?

stwunsch: Or just link the file, without any other comment. Should be clear enough and future safe.

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+- `AsCObject`: equivalent to [cppyy's `as_cobject`](https://cppyy.readthedocs.io/en/latest/lowlevel.html#capsules).
```
stwunsch, line 137: Probably put a half sentence what it does, then it's less cryptic for first comers.

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+- `BindObject`: equivalent to [cppyy's `bind_object`](https://cppyy.readthedocs.io/en/latest/lowlevel.html#c-c-casts).
```
stwunsch, line 139: Same here, half a sentence would be good.

manual/python/index.md
```diff
@@ -60,9 +60,87 @@ caption="Example of graphics generated with PyROOT."
...
+
+```python
+import ROOT
+ROOT.PyConfig.OptionName1 = SomeValue1
+ROOT.PyConfig.OptionName2 = SomeValue2
+```
+
+The available configuration options are described next:
+
+- `DisableRootLogon` (default `False`): just like it happens in C++, PyROOT also supports the [rootlogon](https://root.cern/doc/master/rootlogon_8C.html) functionality. When PyROOT starts, it will look for a file called `.rootlogon.py` in the user's home directory and, if such file exists, PyROOT will import it. This can be used to make some settings every time PyROOT is launched, for example defining some style for your plots:
```
eguiraud, line 79: If I understand correclty `rootlogon.py` is executed at `import ROOT` time (I guess that's what "when PyROOT starts" means)? If so, how can a change of setting that is performed _after_ ROOT is imported change this behavior?

manual/python/index.md
etejedor: Ok!

manual/python/index.md
etejedor: It's `ShutDown`, well spotted, thanks!

manual/python/index.md
etejedor: Jupyter notebook falls in the IPython category, so I can say `interactive mode or IPython/Jupyter` ?

manual/python/index.md
etejedor: ok!

manual/python/index.md
etejedor: I don't see the `adrress` typo?

manual/python/index.md
etejedor: Ok I will add "It is equivalent to..." and ok for the example usage.

manual/python/index.md
etejedor: I think it won't change because if you see the link, it corresponds to a commit. It is the status of the file at that commit.

manual/python/index.md
etejedor: Ok!

manual/python/index.md
etejedor: Ok!

manual/python/index.md
etejedor: There are actually two phases in the PyROOT startup. `import ROOT`only does a few settings, but most of the initialization is done lazily at first lookup (e.g. when you do `ROOT.TTree`). That is why if you configure options after just `import ROOT` it is fine, you are still on time. I can change `When PyROOT starts` by `During PyROOT initialization` if you think that is more clear, but I would not mention the two phases here, I think it will confuse more than help.

eguiraud: ah, I see!

> I think it will confuse more than help.

imho, without mentioning that rootlogon.py is executed upon first lookup and not at `import ROOT` time, this section is very confusing. "during PyROOT initialization" has the same meaning as "when PyROOT starts" for anyone who does not know about the two phases (e.g. me or anyone but you, Stefan, Wim, maybe Axel :D )

manual/python/index.md
etejedor: Mmm I checked and if I do that, I will just copy what is in the cppyy docs, which is only a short sentence. People can just follow the link and read that.

manual/python/index.md
etejedor: I think I explain what the user needs to know: that the options need to be set right after `import ROOT`. The reason why, I don't think its crucial that they know. I would prefer not to go into details about lookups to keep the explanation as simple as possible. In any case, what difference does it make for a user (who knows the configuration options must be set after `import ROOT`) if the real configuration happens at first lookup? Pretty much anything the user will do after `import ROOT` is a lookup anyway.

eguiraud: as you prefer

PR #311, opened 2020-07-02
Link check: survive relative links to non-jekyll pages

### LINKED ISSUE #310: Link checker refuses relative links to non-jekyll pages of root.cern
Linking to `/doc/master` will be signaled as a broken link; it has to be `https://root.cern/doc/master`. That's too bad :-(

Axel-Naumann:
Fixes #310


eguiraud:
I'm pretty sure that is a broken link in presence of a baseurl (e.g. when deploying PR builds), so the tool is right to point it out

PR #312, opened 2020-07-02
Translate old Cling drupal book

couet:


couet:
This is a first draft. All the material for the old drupal book is there.
We surely need to do some cleanup

PR #337, opened 2020-07-21
Update section on serving test websites and PRs.

hageboeck:


hageboeck:
Relevant parts of the preview should be served here:
https://root.cern/337/for_developers/web/#get-the-root-web-site-sources

PR #350, opened 2020-07-24
[Tutorials] Change looks of tutorial links.

hageboeck:
Both Lorenzo and me overlooked the links to the tutorials, because they
are too big. This makes them look less like an icon, and more like a
link.

Old | New
----|-----
image | image

### DIFF
--- a/_includes/tutorials
+++ b/_includes/tutorials
@@ -1,11 +1,9 @@
 {% assign name = {{include.name}}  %}
 {% assign url  = {{include.url}}  %}
 
-<center>
-<a href="https://root.cern/doc/master/group__tutorial__{{url}}.html" target="_blank">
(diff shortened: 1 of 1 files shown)


eguiraud:
I don't really have an opinion, sorry :)

eguiraud:
Ah, note however that the style should be consistent for all manual pages

hageboeck:
It is already consistent, because there's a file that's taking care of generating the links.

couet:
For me it is fine... but that's really a matter of taste and I do no have a strong opinion about it. This tutorial icon/link was initially requested by Johannes. Let me add him in the loop to see what he thinks.
We can also merge this PR and see if we get comments about it ....

eguiraud:
@JohannesHentrich please let us know in case you disagree with this change!

PR #357, opened 2020-07-29
Put "Under the Spotlight" under the spotlight

eguiraud:
Old:

old

New:

new

### DIFF
--- a/index.html
+++ b/index.html
@@ -55,8 +55,9 @@ feature_row:
             <br>
             <div style="padding:0.1em 0.6em;">
                {% for post in site.posts limit:3 %}
-               <font size="2"><b><a href="{{ post.url | relative_url }}">{{ post.title }}</a></b></font> <font size="1">({{post.date | date_to_string}})</font>
-               <font size="2"><p>{{ post.excerpt }}</p></font>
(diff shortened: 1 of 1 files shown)


Axel-Naumann:
I'm not in love. I prefer an inconsistent font size that shows that this is front matter, rather than a small, body-style size.

eguiraud:
The problem I'm trying to solve: blog post titles are currently so small that do not catch the eye at all.

Would you like me to try something else, or would you just leave it as it is?

Axel-Naumann:
Thanks, I agree that's an issue. What about this:
- we merge tech blog and "under the spotlight" - as a use, why do I care about these two being separate? What makes something show up in one instead of the other?
- we remove the header "Under the Spotlight"
- the "under the spotlight" entry titles are sized to match the current "Under the Spotlight" header.

@couet thoughts (esp on separating "spotlight" and "tech blog")?

couet:
The separation between "spotlight" and "tech blog" came from the old web site. I think "spotlight" was for the more recent posts ... that's it ... I agree they can be merged.

eguiraud:
I will not have time to try out those changes before a while, so I converted this discussion to an issue, #362 . Closing this PR.

PR #358, opened 2020-07-29
As per Danilo's request: remove him from about/team

eguiraud:


couet:
May be you should also remove the corresponding picture: /assets/images/DP.jpg

couet:
And also, may be move him to: https://root.cern/about/previous_developers/

eguiraud:
@couet thanks, did not remember we had that section. How about now?

eguiraud:
...i'm guessing it's ok :D

Issue #34, opened 2020-06-08
"News" should include release announcement.


pcanal:

Axel-Naumann:
The front page's "Latest Releases" should at least show in the left side-bar of https://root-project.github.io/web/news/ - or the items should show in the list of items of https://root-project.github.io/web/news/

Axel-Naumann:
And we should consider writing blog / news posts for new major releases; that also takes care of this to some extent.

couet:
This issue might be considered invalid now that /news does not exist anymore.

PR #375, opened 2020-08-05
merge releases and blog side bar

couet:
The releases side bar had only one entry. 
This PR merge it with the blog side which can be also consider as "Useful links" when people
are looking for new about ROOT. This is an attempt to answer this issue: https://github.com/root-project/web/issues/34

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -159,12 +159,6 @@ for_developers:
      url: https://root.cern/img/logos/ROOT_Logo/
      target: _blank
 
-releases:
-   - title: Useful links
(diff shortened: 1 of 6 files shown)


eguiraud:
I find it confusing that /blog has a navbar called "releases" and that when looking at releases blog posts and ROOT workshops are advertised in the side bar.

I think #34 can be considered invalid now that /news does not exist anymore.

couet:
@eguiraud 
Yes, it was a try ... may be we should close #34 ... I will cancel this PR for now I guess.

PR #390, opened 2020-08-18
[PyROOT] Correct section about finding headers when loading libraries.

hageboeck:


hageboeck:
Will re-create inside root-project to benefit from website preview.

PR #391, opened 2020-08-18
[PyROOT] Correct section about finding headers when loading libraries.

hageboeck:


hageboeck:
Preview served here:
https://root.cern/391/manual/python/#4-loading-a-library-with-root-dictionaries

Are the italics in the Nota Bene section OK? How to get rid of those??

PR #397, opened 2020-08-27
[PyROOT] Add entry about boolean value of C++ instance proxies

etejedor:
Reported by Walter Lampl.

### DIFF
--- a/manual/python/index.md
+++ b/manual/python/index.md
@@ -597,3 +597,26 @@ modifier methods on the `std::string` objects.
 <class cppyy.gbl.std.string at 0x4cd41b0>
 <class cppyy.gbl.std.string at 0x4cd41b0>
 ```
+
+- When obtaining the boolean value of a C++ instance proxy, both old and new PyROOT return
+`False` when such proxy points to null. On the other hand, when the proxy points to a C++ object,
+old PyROOT just returns `True`, while new PyROOT has slightly modified this behaviour: in new
+cppyy, if `__len__` is available, the result of `__len__` is used to determine truth. This is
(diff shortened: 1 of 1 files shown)


stwunsch (changes requested):
I would rephrase along the lines of the following proposal:

> When obtaining a boolean value of a C++ instance proxy, new cppyy does an extra check and returns `False` if the C++ pointer is null. Otherwise, the normal Python behaviour is followed (see [here](https://docs.python.org/3.8/reference/datamodel.html#object.__bool__)). For example, the following code shows how the insertion of `__len__` changes the boolean value of an instance proxy:

Would it be possible to create an instance with a zero pointer to show that it's `False`? I cannot think of something nice looking, just ugly tricks.

manual/python/index.md
```diff
@@ -597,3 +597,26 @@ modifier methods on the `std::string` objects.
 <class cppyy.gbl.std.string at 0x4cd41b0>
 <class cppyy.gbl.std.string at 0x4cd41b0>
 ```
+
+- When obtaining the boolean value of a C++ instance proxy, both old and new PyROOT return
+`False` when such proxy points to null. On the other hand, when the proxy points to a C++ object,
```
stwunsch, line 602:
```suggestion
`False` when such proxy points to null. On the other hand, when the proxy points to a valid C++ object,
```

stwunsch (approved):
Looks good! One small improvement could be done by pointing out that we return `True` for a **valid** C++ object.

manual/python/index.md
etejedor: Indeed, I thought about this, but then I realized that even if the object is not valid (e.g. it has been already deleted without the proxy knowing it), it could still return `True`.

PR #398, opened 2020-08-28
[Install] Simplify path to LCG nightlies.

hageboeck:
Just had a question in the forum with users ending up in compiler-not-fitting-root land. Also difference between dev3 and dev3python3 unclear. Therefore, decided to suggest sourcing only one lcg view, and explain py2--py3 difference.

### DIFF
--- a/install/nightlies.md
+++ b/install/nightlies.md
@@ -19,9 +19,10 @@ The [usual instructions]({{ '/install/#download-a-pre-compiled-binary-distributi
 If you have access to LCG, as it is the case on LXPLUS, for example, ROOT nightlies can be obtained by sourcing the relevant scripts:
 
 ```
-source /cvmfs/sft.cern.ch/lcg/nightlies/dev3/<DAY>/lcgenv/*/<PLATFORM+Compiler>/lcgenv-env.sh
-source /cvmfs/sft.cern.ch/lcg/nightlies/dev3/<DAY>/ROOT/HEAD/<PLATFORM+Compiler>/bin/thisroot.sh
+source /cvmfs/sft.cern.ch/lcg/views/dev3python3/latest/<Platform+compiler>/setup.sh
 ```
+- For python2, use `lcg/views/dev3/`
+- To get a ROOT version a few days old, replace `latest` with the desired day.
 
 ## Conda package
 


install/nightlies.md
```diff
@@ -19,9 +19,10 @@ The [usual instructions]({{ '/install/#download-a-pre-compiled-binary-distributi
 If you have access to LCG, as it is the case on LXPLUS, for example, ROOT nightlies can be obtained by sourcing the relevant scripts:
 
 ```
-source /cvmfs/sft.cern.ch/lcg/nightlies/dev3/<DAY>/lcgenv/*/<PLATFORM+Compiler>/lcgenv-env.sh
-source /cvmfs/sft.cern.ch/lcg/nightlies/dev3/<DAY>/ROOT/HEAD/<PLATFORM+Compiler>/bin/thisroot.sh
+source /cvmfs/sft.cern.ch/lcg/views/dev3python3/latest/<Platform+compiler>/setup.sh
```
eguiraud, line 22: interesting, so there is:

* `/cvmfs/sft-nightlies.cern.ch/lcg/views/dev3python3` (or dev4)
* `/lcg/views/dev3python3/` (and friends)
* `/cvmfs/sft.cern.ch/lcg/nightlies/dev3/` (and friends)

I'm confused :D I agree in general views are better than non-views, @Axel-Naumann might know why this was not a link to a view in the first place

hageboeck: First and second are identical I think. Third is what the view will set up for you, so you don't have to know about it.

I will admit, though, that before you would set up only compiler and ROOT. Now, you set up *everything*.
I would argue, though, that that's OK if you want to test a nightly. I certainly prefer the one-point setup, that you don't need a `*` in the path, that you don't have to choose the day.

Axel-Naumann:
The view brings in tons of stuff, often breaking users' setup, e.g. emacs doesn't start after sourcing the view. I am not sure this is an improvement. Not sure as in: if you believe it's worth the risk / downsides let's merge this!

eguiraud:
I don't really have an opinion, sorry :sweat_smile: I always use views because i only use lcg releases on lxplus

hageboeck:
I also only ever use views, because they are easier to set up. :-)

PR #403, opened 2020-09-03
[PyROOT] Add entry about buffers that point to null

etejedor:
From ticket:

https://sft.its.cern.ch/jira/browse/ROOT-11028

### DIFF
--- a/manual/python/index.md
+++ b/manual/python/index.md
@@ -620,3 +620,23 @@ True
 >>> bool(a)  # False because len(a) == 0
 False
 ```
+
+- In new cppyy, buffer objects are represented by the `LowLevelView` type. If such a buffer
+points to null, it is not iterable, unlike in the old PyROOT. For example, in the following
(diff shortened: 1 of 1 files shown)


manual/python/index.md
```diff
@@ -620,3 +620,23 @@ True
...
+>> g = ROOT.TGraph()
+>> x = g.GetX()
+
+>> xl = list(x)  # throws ReferenceError
+```
+
+The code above can be protected by checking first for the validity of `x`:
+
+```python
+>>> if x: xl = list(x)
```
stwunsch, line 641: What about the following?

```suggestion
>>> xl = list(x) if x else 'your default'
```

stwunsch (approved):
I've added two suggestions, but nothing critical. Looks good!

manual/python/index.md
etejedor: Looks good thanks!

etejedor:
Applied @stwunsch comments, thanks!

PR #430, opened 2020-10-06
missing dash

couet:
This missing dash made the item "topical manual" disappeared (thanks Johannes to have see it)


JohannesHentrich (approved):
looks fine

PR #436, opened 2020-10-09
Link to extra Genvector doc

couet:
Link to extra Genvector doc in topical manuals


couet:
We have a problem with the PR. The link to the pdf is "http" and the root web site accept only "https". @lmoneta do you think you can put this pdf on our server ?

lmoneta:
Hi Olivier, 

Yes you should put the pdf in the root.cern.ch server. I am not sure how log seal.cern.ch will be still valid.
Thank you for doing this

Lorenzo

JohannesHentrich (approved):
Very good. Thank you for doing this !

couet:
Hi Lorenzo,

You do not need to log in to seal. It is a pdf file. It is enough to download it from the browser. Here it is:
[GenVector.pdf](https://github.com/root-project/web/files/5353908/GenVector.pdf)
Now I do not know how to put it on root.cern.ch file server.

Cheers,
Oilvier

lmoneta:
I don't have access. But you can add the file also on git (root git or the root web git)

Lorenzo

couet:
Ok done. the pdf file is in the "topical" folder

PR #446, opened 2020-10-15
Editing

JohannesHentrich:
Editing manual/python/index.md
- Removing passive voice
- Removing numbering from the "Loading user libraries and Just-In-Time compilation (JITting)" subsections. The single subchapters are not consecutive and can be read independently.
- Improving readability (hopefully doing so).

### DIFF
--- a/manual/python/index.md
+++ b/manual/python/index.md
@@ -7,13 +7,13 @@ toc: true
 toc_sticky: true
 ---
 
-With PyROOT you can use ROOT, because of its Python-C++ bindings, from Python. PyROOT is HEP's entrance to all C++ from Python, for example, for frameworks and their steering code. The PyROOT bindings are *automatic* and *dynamic*: no pre-generation of Python wrappers is necessary.
+With PyROOT, ROOT's Python-C++ bindings, you can use ROOT from Python. PyROOT is HEP's entrance to all C++ from Python, for example, for frameworks and their steering code. The PyROOT bindings are *automatic* and *dynamic*: no pre-generation of Python wrappers is necessary.
 
-With PyROOT you can acces the full ROOT functionality from Python while benefiting from the performance of the ROOT C++ libraries.
+With PyROOT you can access the full ROOT functionality from Python while benefiting from the performance of the ROOT C++ libraries.
 
 PyROOT is compatible with both Python2 (>= 2.7) and Python3.
 
-> The usage of PyROOT requires in-depth knowledge of Python.<br/>
+> The usage of PyROOT requires working knowledge of Python.<br/>
 > For detailed information on Python, refer to the [Python Language Reference](https://docs.python.org/3/reference/){:target="_blank"}.
 
 Together with ROOT 6.22, a major revision of PyROOT has been released. The new PyROOT has extensive support for *modern C++* (it operates on top of [cppyy](https://cppyy.readthedocs.io/){:target="_blank"}), is more *pythonic* and is able to *interoperate* with widely-used Python data-science libraries (for example, [NumPy](https://numpy.org/){:target="_blank"}, [pandas](https://pandas.pydata.org/){:target="_blank"}, [Numba](https://numba.pydata.org/){:target="_blank"}).<br/>Therefore, we strongly recommend to use the new PyROOT.
@@ -96,13 +96,13 @@ ROOT.gROOT.SetStyle('MyStyle')
 
 If PyROOT cannot find `.rootlogon.py` in the user's home directory, it looks for the equivalent in C++ (`.rootlogon.C`), first in [ROOT's etc directory](https://root.cern/doc/master/classTROOT.html#ab8e51627a12d886d6c8177b46481352a){:target="_blank"}, then in the user's home directory and finally in the current working directory. 
 
-Note that it is also possible to use both the Python and the C++ rootlogons, since the latter can be loaded from the former, for instance with `ROOT.gROOT.LoadMacro('.rootlogon.C')`.
+Note that it is also possible to use both the Python and the C++ rootlogons, since the latter can be loaded from the former, for example with `ROOT.gROOT.LoadMacro('.rootlogon.C')`.
 
 If you want to disable the rootlogon functionality, set `PyConfig.DisableRootLogon` to `True`.
 
 - `IgnoreCommandLineOptions` (default `True`): If a PyROOT script is run with some command line arguments, ROOT ignores them by default, so you can process them as you wish. However, by setting `PyConfig.IgnoreCommandLineOptions` to `False`, those arguments are forwarded to ROOT for parsing, for example, to enable the batch mode from the command line.<br/>For a complete list of the arguments accepted by ROOT, → see [Starting ROOT with command line options]({{ '/manual/first_steps_with_root/#starting-root-with-command-line-options' | relative_url }}).
 
-- `ShutDown` (default `True`): If PyROOT is terminated during its cleanup phase, the ROOT C++ interpreter is shut down.<br/>If PyROOT is executed as part of a longer-running application that needs the C++ interpreter, you can set `PyConfig.ShutDown` to `False` to prevent that shutdown.
+- `ShutDown` (default `True`): When the application is finished, more precisely during PyROOT's cleanup phase, the ROOT C++ interpreter is shut down by default.<br/>If PyROOT is executed as part of a longer-running application that needs the C++ interpreter, you can set `PyConfig.ShutDown` to `False` to prevent that shutdown.
 
 - `StartGUIThread` (default `True`): Unless executed from [IPython](https://ipython.org/){:target="_blank"}, a [Jupyter](https://jupyter.org/){:target="_blank"} notebook or in [interactive mode](https://docs.python.org/3/tutorial/interpreter.html#interactive-mode){:target="_blank"}, PyROOT starts a thread that periodically polls for ROOT events (for example GUI events) to process them. If a given PyROOT application does not need this event processing, you can prevent the creation of the thread by setting `PyConfig.StartGUIThread` to `False`.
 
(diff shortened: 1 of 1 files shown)


couet:
Hi @JohannesHentrich it looks fine for me. As there is quite a lot of changes, I guess we should wait @etejedor approval
 before merging.

couet (approved):

manual/python/index.md
```diff
@@ -7,13 +7,14 @@ toc: true
 toc_sticky: true
 ---
 
-ROOT can be used from Python thanks to its Python-C++ bindings, called PyROOT. PyROOT is HEP's entrance to all C++ from Python, for example, for frameworks and their steering code. The PyROOT bindings are *automatic* and *dynamic*: no pre-generation of Python wrappers is necessary.
+With PyROOT you can use ROOT, because of its Python-C++ bindings, from Python. PyROOT is HEP's entrance to all C++ from Python, for example, for frameworks and their steering code. The PyROOT bindings are *automatic* and *dynamic*: no pre-generation of Python wrappers is necessary.
```
etejedor, line 10: Here I would like to make it clear that PyROOT is the Python bindings, so perhaps something like `With PyROOT, ROOT's Python-C++ bindings, you can use ROOT from Python`?

manual/python/index.md
```diff
@@ -7,13 +7,14 @@ toc: true
...
 
-With PyROOT, all the ROOT functionality can be accessed from Python while, at the same time, benefiting from the performance of the ROOT C++ libraries.
+With PyROOT you can acces the full ROOT functionality from Python while benefiting from the performance of the ROOT C++ libraries.
 
 PyROOT is compatible with both Python2 (>= 2.7) and Python3.
 
-> Using PyROOT requires working knowledge of Python. Detailed information about the Python language can be found in the [Python Language Reference](https://docs.python.org/3/reference/){:target="_blank"}.
+> The usage of PyROOT requires in-depth knowledge of Python.<br/>
```
etejedor, line 16: I wouldn't say `in-depth knowledge` is required, at least for the most common operations. I think `working knowledge` is fine or alternatively `fair knowledge`.

manual/python/index.md
```diff
@@ -81,31 +84,33 @@ ROOT.PyConfig.IgnoreCommandLineOptions = False
...
 
-- `StartGUIThread` (default `True`): except when executed from [IPython](https://ipython.org/){:target="_blank"}, a [Jupyter](https://jupyter.org/){:target="_blank"} notebook or in [interactive mode](https://docs.python.org/3/tutorial/interpreter.html#interactive-mode){:target="_blank"}, PyROOT starts a thread that periodically polls for ROOT events (for example GUI events) to process them. If a given PyROOT application does not need this event processing, you can prevent the creation of the thread by setting `PyConfig.StartGUIThread` to `False`.
+- `ShutDown` (default `True`): If PyROOT is terminated during its cleanup phase, the ROOT C++ interpreter is shut down.<br/>If PyROOT is executed as part of a longer-running application that needs the C++ interpreter, you can set `PyConfig.ShutDown` to `False` to prevent that shutdown.
```
etejedor, line 105: I like it better how it was before, `If PyROOT is terminated during its cleanup phase` does not say what I meant. What about this: `When the application finishes, during PyROOT's cleanup phase, the ROOT C++ interpreter is shut down by default.`.

manual/python/index.md
```diff
@@ -144,29 +151,29 @@ Moreover, you can use `addressof` in conjunction with `TTree::Branch` from Pytho
...
+- `SetOwnership(obj, python_owns)`: A Python proxy can either own or not own its internal C++ object. If a Python proxy owns its C++ object and the proxy is being destroyed, the C++ object is deleted too. This ownership can be modified for a given Python proxy with `SetOwnership`.<br>For example, when calling `SetOwnership(obj, False)`, make sure that the C++ object proxied by `obj` is not deleted by PyROOT when `obj` is garbage collected. This is useful for example, if you know that the deletion will happen in C++ and you want to prevent a double delete.<br/>Use this functionality with care not to produce any memory leaks.
```
etejedor, line 156: In this paragraph, what about this modification: "For example, by calling `SetOwnership(obj, False)`, you make sure ...". It is by using `SetOwnership` that you make sure the C++ object is not deleted.

manual/python/index.md
```diff
@@ -144,29 +151,29 @@ Moreover, you can use `addressof` in conjunction with `TTree::Branch` from Pytho
...
 
 ## Loading user libraries and Just-In-Time compilation (JITting)
 
 With PyROOT you can use *any C++ library* from Python, not only the ROOT libraries. This is possible because of the automatic and dynamic bindings between Python and C++ that PyROOT provides.
 Without any prior generation of wrappers, at execution time, PyROOT can load C++ code and call into it.
 
-This allows for writing high-performance C++, compiling it, and using it from Python.
+This enables you to write high performance C++, compile it and use it from Python. 
 The following options are available, ordered by complexity and performance:
```
etejedor, line 164: I think that the fact that the options are ordered by complexity and performance justifies the numbering below. (instead of just bullets).

manual/python/index.md
```diff
@@ -373,15 +387,15 @@ f1.Draw()
 c.Draw() # Necessary to make the graphics show!
 ```
 
-Executing the code above in a cell will make the following interactive canvas appear as output:
+Executing the code above in a cell displays the following interactive canvas in the output:
```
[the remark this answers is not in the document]
etejedor: In Jupyter notebooks, cells have an "output" section below them, where the output is displayed. I think it's fine.

manual/python/index.md
```diff
@@ -407,7 +421,7 @@ cppyy template proxy (internal)
 'int'
 ```
 
-Note that the above code does not affect class templates, which can be instantiated either with parenthesis or square brackets:
+Note that the above code does not affect class templates that can be instantiated either with parenthesis or square brackets:
```
etejedor, line 424: I think the `, which` is what I meant here: I am providing some extra information about class templates, I am not specifying which class templates are not affected.

manual/python/index.md
```diff
@@ -555,8 +570,8 @@ The purpose of these classes was to serve as a base class for Python classes tha
 from the ROOT::Math classes. This allowed to define Python functions that could be used for fitting
 in Fit::Fitter.<br/>
 In the new PyROOT, `TPyMultiGenFunction` and `TPyMultiGradFunction` do not exist anymore, since their
-functionality is automatically provided by new cppyy: when a Python class inherits from a C++ class,
-a wrapper C++ class is automatically generated. That wrapper class will redirect any call from C++
+functionality is automatically provided by new cppyy. When a Python class inherits from a C++ class,
```
etejedor, line 573: With the `:` here I meant to explain how the functionality is automatically provided by new cppyy.

manual/python/index.md
```diff
@@ -615,14 +630,14 @@ modifier methods on the `std::string` objects.
 
 - When obtaining the boolean value of a C++ instance proxy, both old and new PyROOT return
 `False` when such proxy points to null. On the other hand, when the proxy points to a C++ object,
-old PyROOT just returns `True`, while new PyROOT has slightly modified this behaviour: in new
+old PyROOT just returns `True`, while new PyROOT has slightly modified this behaviour. In new
```
etejedor, line 633: Same comment here about the `:`, perhaps it's a language difference? I use `:` to link with a clarification/further explanation of my previous statement.

etejedor (changes requested):
Thanks a lot @JohannesHentrich ! Please see my comments.

manual/python/index.md
```diff
@@ -144,29 +151,29 @@ Moreover, you can use `addressof` in conjunction with `TTree::Branch` from Pytho
...
+- `SetOwnership(obj, python_owns)`: A Python proxy can either own or not own its internal C++ object. If a Python proxy owns its C++ object and the proxy is being destroyed, the C++ object is deleted too. This ownership can be modified for a given Python proxy with `SetOwnership`.<br>For example, when calling `SetOwnership(obj, False)`, make sure that the C++ object proxied by `obj` is not deleted by PyROOT when `obj` is garbage collected. This is useful for example, if you know that the deletion will happen in C++ and you want to prevent a double delete.<br/>Use this functionality with care not to produce any memory leaks.
 
 ## Loading user libraries and Just-In-Time compilation (JITting)
 
 With PyROOT you can use *any C++ library* from Python, not only the ROOT libraries. This is possible because of the automatic and dynamic bindings between Python and C++ that PyROOT provides.
 Without any prior generation of wrappers, at execution time, PyROOT can load C++ code and call into it.
 
-This allows for writing high-performance C++, compiling it, and using it from Python.
+This enables you to write high performance C++, compile it and use it from Python.
```
hageboeck, line 163: `high-performance C++` was the correct spelling.

hageboeck:
> Removing numbering from the "Loading user libraries and Just-In-Time compilation (JITting)" subsections. The single subchapters are not consecutive and can be read independently.

I actually numbered them so people can select one approach from the list of alternatives above, and then jump down and *only* read "their" number. I'm not strongly attached to those numbers, though. Just pointing out that I instinctively did the exact opposite, but with the same intention.

manual/python/index.md
JohannesHentrich: Your proposal is indeed clearer. I think that the term "is terminated" is somehow more technical than "finishes", but may this is only a matter of taste.

manual/python/index.md
JohannesHentrich: sure, but then it a very long sentence. But anyhow "when" must be reflaced by "if".

manual/python/index.md
JohannesHentrich: I know that people like using "which". But in written English "that" is used and I think every proofreader would change "which" to "that" ( and I can't count the times when it was changed on me :-)) But we can let it as is was. But we can let it as is was

manual/python/index.md
JohannesHentrich: An ordered list is only used, if there is a sequence (especially in tasks). Complexity (which means for everybody something else) is no real ordering criteria (in the Technical Documentation).

if you only scan the menu on the left side, it is assumed that there is a sequence to be followed:
   Loading user libraries and Just-In-Time compilation (JITting)
     1. Just-in-time compilation ...
     2. Just-in-time compilation ..
     3. Loading ...
     4. Loading ...

So in this case a numbered list is not correct.

manual/python/index.md
etejedor: Thanks for the explanation, but if I can't use `which`, how can I express addition of information? My point is what is explained here: https://grammar.yourdictionary.com/grammar-rules-and-tips/when-to-use-which-or-that.html . I don't want to be restrictive in what class templates I'm talking about, I want to add extra info about class templates.

couet:
It seems we can merge now, can't we ?

eguiraud:
Standard practice would be to wait for a green light by the reviewers that requested changes

etejedor (approved):
LGTM!

PR #448, opened 2020-10-16
Add explanation of thread-safety in ROOT

jblomer:


manual/thread_safetiness/index.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Thread safetiness
```
pcanal, line 2:
```suggestion
title: Thread safety
```
(google claim safetiness is (for now?) a misspelling)

jblomer: Oh. But the noun would then be rather `thread safeness`?

manual/thread_safetiness/index.md
```diff
@@ -0,0 +1,23 @@
...
+ROOT classes support one of the following thread-safety levels:
+
+- Thread unsafe: objects must not be used concurrently by different threads, even if objects are not shared among threads.
+
+- Conditionally thread safe: objects can be used safely by multiple threads concurrently as long as mutiple threads do not share the same object (i.e. every thread uses a private copy of its objects).  In addition, objects can be shared among threads as long as all threads use only `const` methods of the shared objects.
+
+- Thread safe: objects can safely be shared and used concurrently by multiple threads.
+
+Unless otherwise stated, classes starting with `R` (e.g. `RDataFrame`) are conditionally thread safe.
+Classes starting with `T` (e.g. `TDirectory`) are thread unsafe.
```
pcanal, line 19: Actually TDirectory (and as far as I can all the 'core' I/O classes) are "conditionally thread safe".

To find an example of thread unsafe.  An example of thread unsafe class is TUrl since it has const method that modify its content (via mutable data members) without taking any locks.

manual/thread_safetiness/index.md
```diff
@@ -0,0 +1,23 @@
...
+layout: single
+sidebar:
+  nav: "manual"
+toc: true
+toc_sticky: true
+---
+
+ROOT classes support one of the following thread-safety levels:
+
+- Thread unsafe: objects must not be used concurrently by different threads, even if objects are not shared among threads.
```
pcanal, line 12: @axel @jblomer is there a nomenclature to distinguish between the "objects must not be used concurrently by different threads" case (i.e. they have unprotected mutable data member for example) and the case where there is an underlying singleton (static data member of function static variable) that makes that means that it not even safe to use distinct object in different threads?

manual/thread_safetiness/index.md
pcanal: I have never seen "safeness" used in this  context (but google says it is used albeit "rarely")

manual/thread_safetiness/index.md
pcanal:
```suggestion
Classes starting with `T` (e.g. `TUrl`) should be considered thread unsafe unless otherwise stated.

...

The classes involved in the I/O (e.g. `TDirectory`) are generally conditionally thread safe.
The classes involved in the retrieving of global information (eg. `TROOT`, `TClass`) are generally thread safe.
Most ROOT collections (eg. `TList`, `TObjArray`) are conditionally thread safe and can be turned into thread safe collection by calling `UseRWLock()` (on each given instances).

```

manual/thread_safetiness/index.md
JohannesHentrich: May some general remarks from my side.

I would like to include the content in one of the existing files/chapters, for example in https://root.cern/manual/root_classes_data_types_and_global_variables/
Otherwise the Basics part becomes too large. May we must reorganize or rewrite one of the existing chapters, but the aim should be not to let the first part become too overwhelming.

Avoid passive. In the ROOT Manual we talk to the reader directly.
Example: "...objects must not be used concurrently by different threads": Who is the actor? If the user is the actor, write: "Make sure that objects are not used concurrently by different threads".
If the application is the actor, name the actor and write: "xxx ensures that objects are not used concurrently by different threads"".


If a class is mentioned the first time, link to the appropriate page in the Reference Guide. Olivier creates a nice macro {% include ref class="ClassName" %}, for example {% include ref class="TDirectory" %} links to https://root.cern/doc/master/classTDirectory.html

Do not write "TODO" on the page that should be published. We do not indicate a work on progress in the Manual. By default a manual is always work in progress :-)

eguiraud: fwiw I agree with @pcanal that the term is [thread safety](https://en.wikipedia.org/wiki/Thread_safety)

manual/thread_safetiness/index.md
eguiraud, line 12: I think the usage of "object" here is a bit misleading: "objects must not be used concurrently..." implies _the same instance_ while "even if objects are not shared among threads" implies different per-thread instances.

Also, it's worse/more complicated than that: the paragraph mentions generic "objects" and the reader would assume we are talking about instances of one given type, but I in principle you could have race conditions between instances of different types that happen to access the same global state (and since we did not design against that nor we test for that, we have to assume this happens somewhere).

manual/thread_safetiness/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Conditionally thread safe: objects can be used safely by multiple threads concurrently as long as mutiple threads do not share the same object (i.e. every thread uses a private copy of its objects).  In addition, objects can be shared among threads as long as all threads use only `const` methods of the shared objects.
```
eguiraud, line 14: A suggestion that's just to hint at what I think should be made clearer, not meant to be accepted as is:

```suggestion
- Thread-local instances are safe: different threads can safely use different instances of the given type (i.e. every thread uses a local instance of this type).  In addition, objects can be shared among threads as long as all threads use only `const` methods of the shared objects. Instances of different types can freely be used concurrently.
```

manual/thread_safetiness/index.md
```diff
@@ -0,0 +1,23 @@
...
+
+- Thread safe: objects can safely be shared and used concurrently by multiple threads.
+
+Unless otherwise stated, classes starting with `R` (e.g. `RDataFrame`) are conditionally thread safe.
```
eguiraud, line 18:
```suggestion
Unless otherwise stated, classes starting with `R` (e.g. `RDataFrame`) offer the basic thread-safety guarantees.
```

manual/thread_safetiness/index.md
eguiraud, line 19: maybe:

```suggestion
Types whose name starts with `T` (e.g. `TDirectory`) as well as all globals (such as `gROOT`) are thread unsafe.
```

eguiraud: "generally" thread safe sounds scary

pcanal: TDirectory is conditionally thread safe.  gROOT is thread safe. :)

pcanal: Well ... english is hard :) the "generally" was meant to bind to the list of classes not the thread safetyness :).  I.e. I was not sure enough to say that "All" the classes involved in the retrieving global information were thread safe.

manual/thread_safetiness/index.md
jblomer: Isn't this the difference between `conditionally thread safe` and `thread unsafe`?

Or are you looking for a term for classes that don't use unprotected global statics but do contain thread unsafe `const` methods?

jblomer: @pcanal @Axel-Naumann Speaking of terms, while Wikipedia calls it "conditionally thread safe", I actually did not hear this term before; instead I would have called the concept "thread friendly", which I also find describes the concept better.  What do others think?

jblomer: Good point.

manual/thread_safetiness/index.md
jblomer: We should add this information to the class documentation of `TDirectory` and `gROOT`.

manual/thread_safetiness/index.md
jblomer: @JohannesHentrich Thanks for the comments!  Should be all addressed expect for the place of this section.  I'll leave it to @couet to decide where to put it.

manual/thread_safetiness/index.md
Axel-Naumann: "Thread friendly" isn't a term of art; "conditionally thread safe" is.

manual/thread_safety/index.md
```diff
@@ -0,0 +1,24 @@
...
+You should not use thread unsafe objects concurrently by multiple threads even if every thread uses its own copy of an object and even if the threads use objects of _different_ (thread unsafe) types.
+
+- Conditionally thread safe: you can use conditionally safe objects concurrently by multiple threads as long as the threads do not share the same object (i.e. every thread uses a local instance of its objects).
+In addition, you can share the same conditionally safe object among threads as long as all threads use only `const` methods of the shared object.
+You can freely use instances of _different_ conditionally safe types concurrently in different threads.
+
+- Thread safe: you can freely use thread safe objects concurrently in multiple threads.
+
+Unless otherwise stated, classes starting with `R` (e.g. {% include ref class="RDataFrame" %}) are conditionally thread safe.
+Types whose name starts `T` (e.g. {% include ref class="TUrl" %}) as well as globals (such as `gStyle`) are thread unsafe.
```
pcanal, line 22: One could argue that TStyle is conditionally thread safe (i.e. literally if you do all the initialization and style setup before starting threads there won't be any problems .. [there is no mutable or caches there])

pcanal:
We could use something like:


Most of the core, math and I/O related classes are conditionally thread safe (`TTree`, `TDirectory`, `TFile`, `TH*`, `TMinuit2`) and most of the general infrastructure classes (`TROOT`, `TClass`) are thread safe.  For more detail see the individual class documentation.

(*) True if and only if thread safety is enabled (eg. call to `ROOT::EnableThreadSatefy()`)

jblomer:
> (*) True if and only if thread safety is enabled (eg. call to `ROOT::EnableThreadSatefy()`)

Ah, right, that of course is an important caveat.  Otherwise, I assume all ROOT objects and globals have to be considered unsafe?

pcanal:
> But ROOT::EnableThreadSafety() has no impact on the R... classes?

It is also true for some (thus all for "simplicity") R... classes, eg. RDataFrame.

manual/thread_safety/index.md
```diff
@@ -0,0 +1,27 @@
...
+sidebar:
+  nav: "manual"
+toc: true
+toc_sticky: true
+---
+
+ROOT classes support one of the following thread safety levels:
+
+- Thread unsafe: make sure that any usage of thread unsafe objects is serialized.
+You should not use thread unsafe objects concurrently by multiple threads even if every thread uses its own copy of an object and even if the threads use objects of _different_ (thread unsafe) types.
```
pcanal, line 13: What do you mean/refer-to with:

> even if the threads use objects of _different_ (thread unsafe) types.

jblomer: It picks up Enrico's comment: if thread A uses an object of type TX and thread B uses one of type TY, they might still access the same global state (e.g. one inherited from the same base class).

pcanal, line 13:
```suggestion
You should not use thread unsafe objects concurrently by multiple threads even if every thread uses its own copy of an object and even if the threads use objects of _different_ (thread unsafe) types (as they may directly or indirectly share state in a thread unsafe manner).
```

jblomer:
@couet I'm not quite sure why the automatic check fails. Should I merge despite of the failure?

couet:
let me check

couet:
@jblomer you have two wrong links:
```
https://root.cern/doc/master/classRDataFrame.html 
https://root.cern/doc/master/classTMinuit2.html 
```
That should be fixed before you merge.

jblomer:
@couet I see, thanks!  The `TMinuit` link I could fix, the `RDataFrame` one might be related to a problem with the `{% include ref class="<CLASSNAME>" %}` macro.  Is the macro expected to work on namespace'd classes, too?

couet:
@jblomer this macro is a shortcut relying on the fact a class has the following URL in doxygen (see for TGraph):
https://root.cern/doc/master/classTGraph.html. 
Because of the ROOT workspace RDataFrame URL is: 
https://root.cern/doc/master/classROOT_1_1RDataFrame.html
In that case you cannot use this macro but you should put the direct Link. I will try to improve this macro to handle workspaces. Meanwhile change your code to a direct link.

couet:
Now ok. You can merge.

PR #453, opened 2020-10-26
User application

JohannesHentrich:
Proposal for the "Creating a user application with ROOT" page


manual/creating_a_user_application/index.md
```diff
@@ -0,0 +1,125 @@
...
+{% highlight C++ %}
+#include <iostream>
+#include "TF1.h"
+#include "TApplication.h"
+#include "TCanvas.h"
+#include "TRootCanvas.h"
+
+int main()
+{
+   TApplication app("app", nullptr, nullptr);
```
bellenot, line 60: I would prefer this (more standard) format:
```suggestion
int main(int argc, char **argv)
{
   TApplication app("app", &argc, argv);
```

manual/creating_a_user_application/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+{% highlight C++ %}
+#include <iostream>
+#include "TF1.h"
+#include "TRint.h"
+#include "TCanvas.h"
+
+int main()
+{
+   TRint app("app", nullptr, nullptr);
```
bellenot, line 100: Same here:
```suggestion
int main(int argc, char **argv)
{
   TRint app("app", &argc, argv);
```

PR #467, opened 2020-11-30
Update "Integrating ROOT into cmake projects"

eguiraud:
Preview at https://root.cern/467/manual/integrate_root_into_my_cmake_project .

### DIFF
--- a/manual/integrate_root_into_my_cmake_project/index.md
+++ b/manual/integrate_root_into_my_cmake_project/index.md
@@ -66,13 +66,13 @@ The following is an example of a project that creates a library including a dict
 
 {% highlight C++ %}
 # CMakeLists.txt for the "event" package. It creates a library with a dictionary and a main program.
-cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
-project(event)
-
 # If ROOT is not installed in a default system location you need to tell CMake where to find it.
 # Sourcing `thisroot.sh` already sets the required environment variables.
-# Otherwise, adding the installation prefix of ROOT to CMAKE_PREFIX_PATH tells cmake where to look:
-#list(APPEND CMAKE_PREFIX_PATH "/path/to/root/installation")
+# Otherwise, you'll have to tell cmake where to look for ROOT,
+# e.g. by passing `-DCMAKE_PREFIX_PATH="/path/to/root/installation` at configuration time
+
+cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+project(event)
 
 # Locate the ROOT package and define a number of useful targets and variables.
 find_package(ROOT REQUIRED COMPONENTS RIO Net)
--- a/manual/integrate_root_into_my_cmake_project/index.md
+++ b/manual/integrate_root_into_my_cmake_project/index.md
@@ -8,8 +8,9 @@ toc_sticky: true
 ---
 
 You can integrate ROOT into a [CMake](https://cmake.org){:target="_blank"} based project.<br/>
-The main interface is the [CMake](https://cmake.org){:target="_blank"} command [find_package(...)](https://cmake.org/cmake/help/latest/command/find_package.html){:target="_blank"}, which
-defines the following standard variables:
+The main interface is the [CMake](https://cmake.org){:target="_blank"} command [find_package(...)](https://cmake.org/cmake/help/latest/command/find_package.html){:target="_blank"}.
+
+Calling `find_package(ROOT)` makes the following variables available:
 
 Variable | Type | Description
 ---------|-------|--------------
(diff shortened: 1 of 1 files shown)


manual/integrate_root_into_my_cmake_project/index.md
```diff
@@ -55,36 +60,33 @@ Generate dictionary for a set of header files (convenient wrapper on top of the
...
 project(event)
 
-# You need to tell CMake where to find the ROOT installation. This can be done in a number of ways:
-#   - ROOT built with classic configure/make use the provided $ROOTSYS/etc/cmake/FindROOT.cmake.
-#   - ROOT built with CMake. Add in CMAKE_PREFIX_PATH the installation prefix for ROOT.
-list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
+# If ROOT is not installed in a default system location you need to tell CMake where to find it.
+# Sourcing `thisroot.sh` already sets the required environment variables.
+# Otherwise, adding the installation prefix of ROOT to CMAKE_PREFIX_PATH tells cmake where to look:
+#list(APPEND CMAKE_PREFIX_PATH "/path/to/root/installation")
```
henryiii, line 75: Isn't a better suggestion to add `-DROOT_DIR="/path/to/root/installation"` to the command line invocation, rather than suggesting it be hardcoded it into a CMakeLists?

manual/integrate_root_into_my_cmake_project/index.md
```diff
@@ -55,36 +60,33 @@ Generate dictionary for a set of header files (convenient wrapper on top of the
...
 
-# Locates the ROOT package and defines a number of variables (e.g. ROOT_INCLUDE_DIRS).
+# Locate the ROOT package and define a number of useful variables (e.g. ROOT_INCLUDE_DIRS).
```
henryiii, line 77:
```suggestion
# Locate the ROOT package and define a number of useful targets and variables.
```

Using the variables is asking for trouble, so better to avoid suggesting an explicit name here, I think.

henryiii (approved):
Minor comments, but looks better :)

manual/integrate_root_into_my_cmake_project/index.md
amadio: I'd say even better would be to suggest `-DCMAKE_PREFIX_PATH=...`  rather than `-DROOT_DIR=...`, as that's a generic way that works for all packages.

eguiraud: how about now?

manual/integrate_root_into_my_cmake_project/index.md
```diff
@@ -55,36 +60,33 @@ Generate dictionary for a set of header files (convenient wrapper on top of the
...
 project(event)
 
-# You need to tell CMake where to find the ROOT installation. This can be done in a number of ways:
-#   - ROOT built with classic configure/make use the provided $ROOTSYS/etc/cmake/FindROOT.cmake.
-#   - ROOT built with CMake. Add in CMAKE_PREFIX_PATH the installation prefix for ROOT.
-list(APPEND CMAKE_PREFIX_PATH $ENV{ROOTSYS})
-
-# Locates the ROOT package and defines a number of variables (e.g. ROOT_INCLUDE_DIRS).
+# Locate the ROOT package and define a number of useful targets and variables.
 find_package(ROOT REQUIRED COMPONENTS RIO Net)
```
oshadura, line 78: Let's add here as well `Tree` to make it match to list of `target_link_libraries`?

oshadura (approved):

manual/integrate_root_into_my_cmake_project/index.md
eguiraud: or remove `ROOT::Tree` from the list above, I don't know why I added it

manual/integrate_root_into_my_cmake_project/index.md
```diff
@@ -55,36 +60,33 @@ Generate dictionary for a set of header files (convenient wrapper on top of the
...
-# Defines useful ROOT functions and ROOT macros (e.g. ROOT_GENERATE_DICTIONARY).
-include(${ROOT_USE_FILE})
-
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 ROOT_GENERATE_DICTIONARY(G__Event Event.h LINKDEF EventLinkDef.h)
 
 # Creates a shared library with a generated dictionary.
 add_library(Event SHARED Event.cxx G__Event.cxx)
-target_link_libraries(Event ${ROOT_LIBRARIES})
+target_link_libraries(Event PUBLIC ROOT::RIO ROOT::Net ROOT::Tree)
```
eguiraud, line 85:
```suggestion
target_link_libraries(Event PUBLIC ROOT::RIO ROOT::Net)
```

manual/integrate_root_into_my_cmake_project/index.md
eguiraud: (done)

eguiraud:
I believe I have addressed all comments, will merge later today if nothing else comes up. Thanks to all reviewers!

manual/integrate_root_into_my_cmake_project/index.md
henryiii: Now you overwrite `CMAKE_PREFIX_PATH` with a cached variable so the CMakeLists.txt can't set it. While it's not the best practice, some CMakeLists do mess with it - it would be better to be able to set the directory per package in the examples for those packages, and `<package_name>_DIR` is generic and works for all packages.

eguiraud: if I understand correctly the discussion is now about `-DCMAKE_PREFIX_PATH` vs `-DROOT_DIR`. I don't really know enough to have a preference, @amadio @oshadura what do you say?

amadio: I think Henri's argument is good. Go for `-DROOT_DIR=...` then.

eguiraud: alright, `-DROOT_DIR` it is then! done!

PR #484, opened 2021-01-19
Add Snap package to /install

eguiraud:
@MrCarroll let me know if you would have anything to add. This will end up below https://root.cern/install#conda

### DIFF
--- a/install/index.md
+++ b/install/index.md
@@ -53,6 +53,19 @@ More instructions about using the conda package are available in [this blog post
 
 Please report any issues with the conda package [here](https://github.com/conda-forge/root-feedstock){:target="\_blank"}.
 
+## Snap
+
+On many Linux distributions, ROOT can be installed via Snap. For example, on Ubuntu:
+
+```sh
+$ sudo snap install root-framework
+$ snap run root-framework
+# or if there is no fear of conflicts with other installations:
+$ root # and the output of `which root` should contain `/snap`
+```
+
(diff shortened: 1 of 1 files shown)


MrCarroll:
Most other ROOT installations have `thisroot.sh` or `conda activate` or similar to setup the `$PATH` properly. Since the Snap doesn't I think there's the possibility for confusion with users, particularly, snap usually appends to the end of $PATH and the other installs usually prepend to the front of $PATH, giving them them priority when invoked, and from what I see often people do this in their Bash login files.

I think there might be value in recommending people to use `which root` to ensure the environment is working as they expect, and possibly mention `snap run root-framework` as an explicit way to get the snap regardless of what has priority in the $PATH. If `which root` returns something with `/snap` in the string, you can probably assume it's actually working.

eguiraud:
@MrCarroll is this better?

eguiraud:
A preview of these changes is available at https://root.cern/484/install/#snap

MrCarroll:
Looks good to me!

The `snap run` version has one specific advantage beyond being explicit, it also works in the event the user skipped rebooting the system when installing snapd, which would mean the snap binary is available but the snaps themselves aren't yet in $PATH, it rarely comes up, but inevitably it'll happen eventually so ;)

eguiraud:
@couet @Axel-Naumann I'll merge later today if you don't have anything against it.

Issue #431, opened 2020-10-07
Ubuntu dependencies are out of date


eguiraud:
On Ubuntu20.04, trying to execute the commands at https://root.cern/install/dependencies/#ubuntu-and-other-debian-based-distributions ,
I get

```
E: Unable to locate package openssl-dev                                              
E: Package 'libqt4-dev' has no installation candidate 
```

`openssl-dev` is substituted by `libssl-dev`, while I could not find a substitute package for `libqt4-dev` (there is no `libqt-dev` or `libqt5-dev` as far as I can tell).

eguiraud:
#522 addresses the libssl-related part of this issue.

eguiraud:
When using just the required packages, xrootd is turned on but cannot be built because of missing `uuid`, and the build system also tries to build LZMA but it fails because it's missing the `aclocal` command (this last one is resolved by installing `autotools-dev` and/or `automake`)

ferdymercury:
For Ubuntu 22 LTS and ROOT latest stable, this issue would be fixed by:  https://github.com/root-project/web/pull/936 and https://github.com/root-project/web/pull/935

PR #522, opened 2021-03-09
Replace openssl-dev fully with libssl-dev in section on dependencies on Ubuntu and Debian

bernhardmgruber:
I just tried to build ROOT master today and installed the required dependencies as written here into my Ubuntu 20.10 WSL: https://root.cern/install/dependencies/#ubuntu-and-other-debian-based-distributions. It states there:

> `sudo apt-get install dpkg-dev cmake g++ gcc binutils libx11-dev libxpm-dev \`
> `libxft-dev libxext-dev python openssl-dev`
> On Debian, substitute openssl-dev with libssl-dev.

I also needed `libssl-dev` on my Ubuntu. So I propose to update the documentation.

Furthermore, I seems `libssl-dev` is available since Ubuntu 16.04: https://packages.ubuntu.com/xenial/libssl-dev. On the contrary, there is no direct search hit for `openssl-dev`: https://packages.ubuntu.com/search?keywords=openssl-dev&searchon=names

Could any other Ubuntu expert verify that `libssl-dev` is the correct package?

### DIFF
--- a/install/dependencies.md
+++ b/install/dependencies.md
@@ -85,16 +85,14 @@ Use `sudo apt-get install <package>` or use the graphical "Synaptic Package Mana
 *   **libpng:** png library
 *   **libjpeg:** jpeg library
 *   **python:** (ROOT6 requires version >= 2.7)
-*   **openssl-dev** or **libssl-dev:** for /usr/include/openssl/pem.h and /usr/lib/libssl.so and /usr/lib/libcrypto.so
+*   **libssl-dev:** for `/usr/include/openssl/pem.h`, `/usr/lib/libssl.so` and `/usr/lib/libcrypto.so` (on older systems the package might be called `openssl-dev`)
 
 As a one-liner:
 
 ```bash
 sudo apt-get install dpkg-dev cmake g++ gcc binutils libx11-dev libxpm-dev \
-libxft-dev libxext-dev python openssl-dev
+libxft-dev libxext-dev python libssl-dev
 ```
-
-On Debian, substitute `openssl-dev` with `libssl-dev`.
 
 ### Most common optional packages
 


install/dependencies.md
```diff
@@ -85,17 +85,15 @@ Use `sudo apt-get install <package>` or use the graphical "Synaptic Package Mana
 *   **libpng:** png library
 *   **libjpeg:** jpeg library
 *   **python:** (ROOT6 requires version >= 2.7)
-*   **openssl-dev** or **libssl-dev:** for /usr/include/openssl/pem.h and /usr/lib/libssl.so and /usr/lib/libcrypto.so
+*   **libssl-dev:** for /usr/include/openssl/pem.h and /usr/lib/libssl.so and /usr/lib/libcrypto.so
```
eguiraud, line 88:
```suggestion
*   **libssl-dev:** for `/usr/include/openssl/pem.h`, `/usr/lib/libssl.so` and `/usr/lib/libcrypto.so` (on older systems the package might be called `openssl-dev`)
```

eguiraud (changes requested):
Thank you @bernhardmgruber !

Corresponding issue: https://github.com/root-project/web/issues/431

We are pretty bad at keeping these dependency lists up to date unfortunately -- we would need to automatically validate them in Docker containers for the different Linux flavors and OS versions that we want to support, and that needs someone to set up the jobs.

That's just to clarify the situation. For the PR: `openssl-dev` was the correct package _at some point_, so I think we should leave a trace of it even if we switch to `libssl-dev` as the suggested package (see e.g. my suggestion).

install/dependencies.md
bernhardmgruber: Thank you for the great explanation and for finding the corresponding issue! It's probably a safer bet to still mention `openssl-dev`.

I amended my commit to include you change!

eguiraud (approved):

PR #527, opened 2021-03-13
Adjust font size of `ROOT File Viewer`'s demo GIF

AlbertoPdRF:
How does this look @Axel-Naumann? I can try with a bit bigger font if you think it would be better!

ROOT File Viewer demo GIF

EDIT: added 3D view and used a slightly bigger font size


Axel-Naumann (approved):
Wonderful, thank you!

PR #551, opened 2021-03-31
Update dependencies.md

petrstepanov:
1. Provided correct Fedora 33 names for following libraries: python3-numpy, R, R-Rcpp, R-RInside
2. Updated one-liner commands with missing required libraries.
3. Fixed typos.

### DIFF
--- a/install/dependencies.md
+++ b/install/dependencies.md
@@ -64,7 +64,7 @@ sudo yum install redhat-lsb-core gcc-gfortran pcre-devel \
 mesa-libGL-devel mesa-libGLU-devel glew-devel ftgl-devel mysql-devel \
 fftw-devel cfitsio-devel graphviz-devel \
 avahi-compat-libdns_sd-devel openldap-devel python-devel \
-python3-numpy libxml2-devel gsl-devel R R-Rcpp R-RInside
+python3-numpy libxml2-devel gsl-devel R-devel R-Rcpp-devel R-RInside-devel
 ```
 
 ## Ubuntu and other Debian-based distributions
--- a/install/dependencies.md
+++ b/install/dependencies.md
@@ -32,8 +32,8 @@ Use `yum install <package>` or the graphical "Add/Remove Software" program.
 As a one-liner:
 
 ```bash
-sudo yum install git cmake3 gcc-c++ gcc binutils \
-libX11-devel libXpm-devel libXft-devel libXext-devel openssl-devel
+sudo yum install git make cmake3 gcc-c++ gcc binutils \
+libX11-devel libXpm-devel libXft-devel libXext-devel python openssl-devel
 ```
 
 ### Most common optional packages
@@ -41,9 +41,9 @@ libX11-devel libXpm-devel libXft-devel libXext-devel openssl-devel
 *   **redhat-lsb-core:** needed by some tests using lsb_release
 *   **gcc-gfortran:** for /usr/bin/gfortran
 *   **pcre-devel:** for /usr/bin/pcre-config
-*   **mesa-libGL-deve**l: for /usr/include/GL/gl.h and for /usr/lib[64]/libGL.so
-*   **mesa-libGLU-devel**: for /usr/include/GL/glu.h and for /usr/lib[64]/libGLU.so
-*   **glew-devel (may need enabling of the [EPEL](https://fedoraproject.org/wiki/EPEL) additional software repository)**: for /usr/include/GL/glew.h and for /usr/lib[64]/libGLEW.so
+*   **mesa-libGL-devel:** for /usr/include/GL/gl.h and for /usr/lib[64]/libGL.so
+*   **mesa-libGLU-devel:** for /usr/include/GL/glu.h and for /usr/lib[64]/libGLU.so
+*   **glew-devel (may need enabling of the [EPEL](https://fedoraproject.org/wiki/EPEL) additional software repository):** for /usr/include/GL/glew.h and for /usr/lib[64]/libGLEW.so
 *   **ftgl-devel:** for /usr/bin/pkg-config
 *   **mysql-devel:** for /usr/bin/mysql_config
 *   **fftw-devel:** for /usr/include/fftw3.h and for /usr/lib/libfftw3.so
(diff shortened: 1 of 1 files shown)


eguiraud:
Thanks @petrstepanov ! Would you be able to tell whether the update suggested commands also work on CentOS? Otherwise I can check.

petrstepanov:
> Would you be able to tell whether the update suggested commands also work on CentOS?

Good point. I've only tested Fedora 33 clean install. Indeed, Scientific Linux and CentOS tests need to be carried out as well (heading says "Fedora, Scientific Linux and CentOS"). These distributions may bring extra dependencies into the list.

I've only tested Fedora and unfortunately I do not have enough time at the moment to test dependencies for other distributions. Can someone please take care of that? I assume CentOS tests would require testing both versions, 7 and 8 since many organizations would stick to v7 and not update for obvious reasons.

eguiraud:
Ok let's do it like this: please split "Fedora, Scientific Linux and CentOS" into "Fedora" and "Scientific Linux and CentOS". We can then apply your updates to "Fedora" and leave "Scientific Linux" and "CentOS" as they are (until we can spare some cycles to verify things). There will be a bit of redundancy but it's better than having a list for 3 distros with some packages that are only needed on one or the other.

petrstepanov:
> split "Fedora, Scientific Linux and CentOS" into "Fedora" and "Scientific Linux and CentOS"

I'm afraid that the Dependencies web page will become somewhat bloated then. Since all these systems use same package names then separate "Fedora" and "Scientific Linux and CentOS" sections will almost duplicate each other. 

> There will be a bit of redundancy but it's better than having a list for 3 distros with some packages that are only needed on one or the other.

I honestly do not see this being an issue. It cannot be the case that a package is only needed on one system. Any dependency required by specific version of the ROOT source code is required to be installed on every system. It's just a matter if it comes pre-installed with the distro or not. From my perspective it does not "hurt" to include extra dependencies in the "yum ..." command, even if they come pre-installed.

Let's keep this PR open, no rush merging it right now. I will test recent Scientific Linux and CentOS distributions and add more commits to this PR.

eguiraud:
I am not sure package names are/will always be exactly the same between Fedora and CentOS. I'm thinking e.g. different conventions around python versions, `cmake` vs `cmake3`, `dev -> devel` migrations etc.

But if you can test at least CentOS 8 and the package list works, that's enough for me to merge this.

petrstepanov:
Hi Enrico, I have sorted out the dependencies for Red Hat family distributions.

You are right, there are discrepancies in Python, NumPy and cmake versions. Older RHEL7 based systems (Centos7 and Scientific Linux 7) have python v2 and cmake v2.8 installed by default. On the other hand RHEL8 and Fedora are apparently shipped with Python v3 and cmake v3.1. I have summarized these details in the document.

Also I have added the instructions on adding extra repositories required by RHEL7 and RHEL8 systems. Namely they both require adding EPEL repository. RHEL8 also [requires enabling PowerTools](https://root-forum.cern.ch/t/root-dependencies-can-not-install-in-centos8/42010) repository.

The dependencies ended up splitting in three groups: "Fedora", RHEL8 based (CentOS8) and RHEL7 based (CentOS7 and Scientific Linux 7). I built ROOT 6.22.08 with  `-Dall=ON` and `-DCMAKE_BUILD_TYPE=DEBUG` on three freshly installed systems in my virtual machine: Fedora 33, CentOS 8.3.2011 and Scientific Linux 7. I did not test CentOS7, but I expect it to work pretty much like Scientific Linux 7.

Please let me know if everything looks good and we can proceed with the merge. Looking forward to hearing from you!

install/dependencies.md
```diff
@@ -18,55 +18,93 @@ Use `yum install <package>` or the graphical "Add/Remove Software" program.
...
+As a one-liner for Fedora 33:
+```bash
+sudo yum install redhat-lsb-core gcc-gfortran pcre-devel \
+mesa-libGL-devel mesa-libGLU-devel glew-devel ftgl-devel mysql-devel \
+fftw-devel cfitsio-devel graphviz-devel libuuid-devel \
+avahi-compat-libdns_sd-devel openldap-devel python3-numpy \
+libxml2-devel gsl-devel readline-devel R-devel R-Rcpp-devel R-RInside-devel
+```
+
+CentOS 8 requores adding EPEL repository. Additionally, some packages in CentOS 8 [were moved to the PowerTools repository](https://root-forum.cern.ch/t/root-dependencies-can-not-install-in-centos8/42010). Therefore it needs to be enabled as well:
```
eguiraud, line 85:
```suggestion
CentOS 8 requires adding the EPEL repository. Additionally, some packages in CentOS 8 [were moved to the PowerTools repository](https://root-forum.cern.ch/t/root-dependencies-can-not-install-in-centos8/42010). Therefore it needs to be enabled as well:
```

eguiraud (approved):
Wow, thank you very much for putting in the work, much appreciated!

This looks good to me. Even if it wasn't perfect (not that I see any particular issue) it's definitely an improvement w.r.t. what we had before.

@Axel-Naumann is back next week and I would like to briefly discuss this with him before merging, but other than that looks ok.

install/dependencies.md
petrstepanov: My only concern is following. I simply compiled ROOT with `-Dall=ON` and `-DCMAKE_BUILD_TYPE=DEBUG`. However I am not sure if above options require the largest possible number of the dependencies. Please let me know.

petrstepanov:
My only concern is following. I simply compiled ROOT with `-Dall=ON` and `-DCMAKE_BUILD_TYPE=DEBUG`. However I am not sure if above options require the largest possible number of the dependencies. Please let me know.

install/dependencies.md
eguiraud: can you add a commit with this and the other minor fixes please?

petrstepanov: done!

eguiraud:
Merging after checking with Axel. Thanks again @petrstepanov !

petrstepanov:
My pleasure!

PR #564, opened 2021-04-21
Added directions for updating/using the `latest-stable` branch

jalopezg-r00t:
The `latest-stable` branch is a convenient way for users that build from source to get
the latest revision that is considered stable by the ROOT team.

The name of this branch does not change, meaning that a `git clone -b latest-stable`
will fetch the the most recent revision without manually specifying a tag name.
Also, this branch is meant to be updated regularly as part of the release
procedure. As such, users can upgrade to the latest published version by
issuing a simple `git pull`.

TL;DR
We attain this by manually committing (via `git commit-tree`) the tree of the
latest stable version to the `latest-stable` branch. Basically, we create a linear
history in which each commit points to the tree of the latest tagged release.

On the client-side, for a `git pull` this is always seen as a fast-forward, and
hence the update is trivial, not requiring a merge.

### DIFF
--- a/for_developers/release_checklist/index.md
+++ b/for_developers/release_checklist/index.md
@@ -85,6 +85,11 @@ This assumes you try to create `v6-22-00-patches`, adjust accordingly.
       - `make` must succeed
   1. Push to GitHub
       - `git push origin v6-22-02`
+  1. Update the stable branch. Users that have cloned this branch will receive updates as a fast-forward via `git pull`
+      - `LATEST_STABLE=v6-xx-yy    # e.g. v6-22-02`
+      - If the stable branch does not exist yet, the `-p refs/heads/latest-stable` part must be removed from the command below.
+      - `$ git update-ref refs/heads/latest-stable $(git commit-tree $LATEST_STABLE^{tree} -p refs/heads/latest-stable -m "Updated 'latest-stable' branch to $LATEST_STABLE")`
+      - `$ git push origin latest-stable`
   1. Produce binary tar-files (optional for development releases and release candidates)
       - Start the procedure [root-release-6.22](https://lcgapp-services.cern.ch/root-jenkins/job/root-release-6.22/){:target="_blank"} (or whichever branch) in Jenkins
   1. Install binaries to CVMFS (optional for development releases and release candidates)
--- a/install/build_from_source.md
+++ b/install/build_from_source.md
@@ -30,6 +30,16 @@ The following are the basic instructions for UNIX-like systems. We will use the 
 # substitute `v6-22-00-patches` with the patches branch of the latest release
 $ git clone --branch v6-22-00-patches https://github.com/root-project/root.git root_src
 ```
+
+If you are interested in the latest stable (as opposed to a specific) version, you can get a copy as follows:
+```bash
+$ git clone --branch latest-stable https://github.com/root-project/root.git root_src
+```
+This branch is expected to be updated regularly on each release. Therefore, users who cloned this branch can upgrade to the latest release in a straightforward way:
+```bash
+$ git pull
+```
+
 In the following we will refer to the directory where ROOT sources are (e.g. `root_src` above) as `<sourcedir>`.
 1. Create a directory for the build and a directory for the installation. It is not supported to build ROOT in the source directory. Then change (`cd`) to the build directory:
 ```bash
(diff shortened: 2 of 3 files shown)


for_developers/release_checklist/index.md
```diff
@@ -85,6 +85,10 @@ This assumes you try to create `v6-22-00-patches`, adjust accordingly.
       - `make` must succeed
   1. Push to GitHub
       - `git push origin v6-22-02`
+  1. Update the stable branch. Users that have cloned this branch will receive updates as a fast-forward via `git pull`
+      - `export LATEST_STABLE=v6-xx-yy    # e.g. v6-22-02`
+      - If the stable branch does not exist yet, the `-p refs/heads/stable` part must be removed from the command below.
+      - `$ git update-ref refs/heads/stable $(git commit-tree $LATEST_STABLE^{tree} -p refs/heads/stable -m "Update stable branch to $LATEST_STABLE")`
```
jalopezg-r00t, line 91: @hahnjo and I discussed about also adding a `-p $LATEST_STABLE`.  That would make the commit appear as a merge.  What do you think?  Does it improve the 'readability' of the history?

for_developers/release_checklist/index.md
```diff
@@ -85,6 +85,10 @@ This assumes you try to create `v6-22-00-patches`, adjust accordingly.
       - `make` must succeed
   1. Push to GitHub
       - `git push origin v6-22-02`
+  1. Update the stable branch. Users that have cloned this branch will receive updates as a fast-forward via `git pull`
+      - `export LATEST_STABLE=v6-xx-yy    # e.g. v6-22-02`
```
eguiraud, line 89: do you need the export?
```suggestion
      - `LATEST_STABLE=v6-xx-yy    # e.g. v6-22-02`
```

for_developers/release_checklist/index.md
jalopezg-r00t: Another thing to discuss: should the first commit in the `stable` branch have no parents (orphan branch).  This is only relevant if the user is allowed to perform any operations that involve any other branch.

install/build_from_source.md
```diff
@@ -30,6 +30,16 @@ The following are the basic instructions for UNIX-like systems. We will use the
...
 ```
+
+If you are interested in the latest stable (as opposed to a specific) version, you can get a copy as follows:
+```bash
+$ git clone --branch stable https://github.com/root-project/root.git root_src
+```
+This branch is expected to be updated regularly on each release. Therefore, users who cloned this branch can upgrade to the latest release in a straightforward way:
+```bash
+$ git pull
+```
```
eguiraud, line 41: I guess we can completely get rid of the command that checks out `v6-22-00-patches` in favor of this one, which works independently of what the latest stable is (yes, `stable` does not include the latest patches, but that's a price I'm glad to pay so we have an `install` documentation that's independent from the ROOT versions)

eguiraud (approved):
LGTM, thanks @jalopezg-r00t , minor comments below

for_developers/release_checklist/index.md
jalopezg-r00t: Ofc, not.  Thanks for spotting!

install/build_from_source.md
jalopezg-r00t: Hmmm...  @Axel-Naumann, +1 or not?

jalopezg-r00t: @Axel-Naumann also suggested a blog post to gain more visibility of this new branch.  However, I would make it in a different PR once the `stable` branch has been created, which is expected for the next release.

for_developers/release_checklist/index.md
hahnjo: I think if we don't go for a merge commit (`-p $LATEST_STABLE^{commit}` Javier mentioned above), I'd prefer an orphan branch - I don't see a reason why the first commit would be special.

install/build_from_source.md
hahnjo: :+1: for doing a blog post, after the `stable` branch is there. But I think we should start the `stable` branch for v6.24/00, right?

hahnjo (approved):

Axel-Naumann:
Yes, we'll create the `stable` branch before the blog post goes live.

Should it be `stable` or `latest`? Because `stable` sounds like it doesn't change?

jalopezg-r00t:
> Yes, we'll create the `stable` branch before the blog post goes live.
> 
> Should it be `stable` or `latest`? Because `stable` sounds like it doesn't change?

I don't have any specific preference. What about combining both terms, i.e. `latest-stable`?

Axel-Naumann:
Pseudo-polling in https://mattermost.web.cern.ch/root/pl/eyx76jira383ff8este4oewb1a

PR #583, opened 2021-05-19
[release_checklist] latest-stable: add also as parent the commit of the latest tagged version

jalopezg-r00t:
Initially, we discussed that the `latest-stable` branch would have a unrelated
history (orphan branch). However, adding as parent the latest tagged version
has two benefits:

- Pushes take less time, as git client is able to determine that no objects should be sent (other than the commit itself).
- Comparing branches à la `git show-branch ...` is also possible.


hahnjo (approved):
:+1: as discussed

PR #611, opened 2021-07-05
Bring "how to use my class with ROOT" up to date.

Axel-Naumann:


eguiraud:
I guess Axel pinged me because he knows I'm opinionated about these manual pages, so here we go :smile: Some of the content you add for PyROOT is already at https://root.cern/manual/python/#loading-user-libraries-and-just-in-time-compilation-jitting . The content of "Adding a class to ROOT" and "ROOT macros and shared libraries" is super confusing (slightly less with this PR).

I would suggest a bigger refactoring. Remove "Adding a class to ROOT" and "ROOT macros and shared libraries" and add the following pages:

- `Reading and writing your C++ classes to/from ROOT files`: explains the difference between reading/writing to TFile and reading/writing to TTree (and probably mentions RNTuple), explains `splitlevel`, the need for dictionaries, the fact that STL classes don't need dictionaries, and a bunch of other stuff that's currently either not explained anywhere or super hard to find
- `Creating ROOT dictionaries`: explains what dictionaries are and do, explains the LinkDef syntax, explains the 3 different ways in which one can generate dictionaries: ACLiC, cmake+`ROOT_GENERATE_DICTIONARY` and via the commandline with `rootcling`

We sorely need these pages, we explain this stuff several times per week on the forum.

couet:
replaced by https://github.com/root-project/web/pull/614 to resolve the conflict. Sorry for that.

Axel-Naumann:
> I would suggest a bigger refactoring.

We'll reserve a week to overhaul the manual etc; details later and through a different channel.

PR #623, opened 2021-07-15
Made Website into Progressive Web App

AdvaitDhingra:
Progressive Web Apps give a native app like experience for websites. When root.cern is opened one can install it and it will run like a native application. 

To achieve this I added a manifest.json file and made a service worker in Javascript.

It passed the Google Lighthouse test:

Screenshot from 2021-07-14 22-32-50

And it is installable on both iOS and Android. 

When one opens root.cern one gets a prompt to install it.


Axel-Naumann:
I suppose that's mostly useful for the reference guide? I know the forum is "installable" already. Can you describe the prupose of this, @AdvaitDhingra ?

AdvaitDhingra:
> I suppose that's mostly useful for the reference guide? I know the forum is "installable" already. Can you describe the prupose of this, @AdvaitDhingra ?

Yes, the forum is already installable. This update would allow the root website to be used (to a certain extent) offline and also installable as a pwa so that you wouldn't always have to open your browser and type roo.cern.

AdvaitDhingra:
Since the reference guide isn't a standalone website I made the whole root.cern a pwa

Axel-Naumann:
But the failures in the HTML proofer tell me that you needed to modify each web page (by changing the header). That would likely mean that the reference guild won't be affected by this PR? (That's not an argument against merging this - it's maybe a request to also add this to [ROOT's doxygen header](https://github.com/root-project/root/blob/master/documentation/doxygen/htmlheader.html) :-)

AdvaitDhingra:
> But the failures in the HTML proofer tell me that you needed to modify each web page (by changing the header). That would likely mean that the reference guild won't be affected by this PR? (That's not an argument against merging this - it's maybe a request to also add this to [ROOT's doxygen header](https://github.com/root-project/root/blob/master/documentation/doxygen/htmlheader.html) :-)

Oh, I didn't know there were more headers 😅😂. I'll add this to the others too 🙂

serviceWorker.js
```diff
@@ -0,0 +1,61 @@
...
+self.addEventListener("install", function (event) {
+  console.log("Install Event processing");
+
+  event.waitUntil(
+    caches.open(CACHE).then(function (cache) {
+      console.log("Cached offline page during install");
+
+      if (offlineFallbackPage === "ToDo-replace-this-name.html") {
+        return cache.add(new Response("Update the value of the offlineFallbackPage constant in the serviceworker."));
+      }
```
jalopezg-r00t, line 15: Do we really need this?  ;-)

serviceWorker.js
```diff
@@ -0,0 +1,61 @@
...
+
+// If any fetch fails, it will look for the request in the cache and serve it from there first
+self.addEventListener("fetch", function (event) {
+  if (event.request.method !== "GET") return;
+
+  event.respondWith(
+    fetch(event.request)
+      .then(function (response) {
+        console.log("Add page to offline cache: " + response.url);
+
```
jalopezg-r00t, line 29: I think we could also remove this.

manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "ROOT: analyzing petabytes of data, scientifically.",
+  "short_name": "ROOT",
+  "description": "ROOT Website",
+  "theme_color": "#346295 ",
+  "background_color": "#346295 ",
```
jalopezg-r00t, line 6: Maybe remove the extra spaces at the end.

jalopezg-r00t (changes requested):
Thanks for the contribution!  Please, consider also the following suggestions.  :-)

serviceWorker.js
AdvaitDhingra: Now that I think about it, not really 😅

serviceWorker.js
```diff
@@ -0,0 +1,61 @@
+const CACHE = "root-offline";
```
Axel-Naumann, line 1: Where does this file come from? Did you write this yourself? If not, what is the license of the file and where does it come from?

Axel-Naumann (changes requested):
Clarify origin

serviceWorker.js
AdvaitDhingra: No it's just a name I came up with of the cache file.

AdvaitDhingra:
> Clarify origin

What exactly do you mean by that, sorry 😅 . Should I clarfiy the directory where manifest.json can be found?

Axel-Naumann:
@AdvaitDhingra sorry for the confusion. I would like to know where the files you are proposing here are coming from: whether you have authored them yourself, whether you adapted them, or whether they are straight copies, for each of the files, and in case where you did not write them completely yourself, what their origin and license is. That's needed to determine whether we can include this. Referencing the origin is also simply good style :-)

AdvaitDhingra:
> @AdvaitDhingra sorry for the confusion. I would like to know where the files you are proposing here are coming from: whether you have authored them yourself, whether you adapted them, or whether they are straight copies, for each of the files, and in case where you did not write them completely yourself, what their origin and license is. That's needed to determine whether we can include this. Referencing the origin is also simply good style :-)

Ah ok. I used online guides but the code used is available by many sources. I can try to find an origin.

Axel-Naumann:
That's fine - if there is no single source that you can point to where you copied a file from, but you have "cooked" this out of multiple online sources, then we don't have a copyright / license question to address.

AdvaitDhingra:
> That's fine - if there is no single source that you can point to where you copied a file from, but you have "cooked" this out of multiple online sources, then we don't have a copyright / license question to address.

Okay, thanks. In the mean time I'll look for a way to link the manifest correctly

Axel-Naumann:
See e.g. https://github.com/root-project/web/blob/7389c44d030bee71d90bfc7f45f5ea2de011a469/_includes/footer.html#L4 for how to create such a link

AdvaitDhingra:
> See e.g.
> 
> https://github.com/root-project/web/blob/7389c44d030bee71d90bfc7f45f5ea2de011a469/_includes/footer.html#L4
> 
> for how to create such a link

Okay, thanks

AdvaitDhingra:
The problem is that the index.html file is in the root directoy, so for that manifest.json is just manifest.json but for the others (refrence for example) its ../manifest.json .

AdvaitDhingra:
That did it. Instead of {{site.baseurl}} I used {{site.url}} . Since baseurl was static and url is just root.cern

AdvaitDhingra:
@Axel-Naumann is everything good?

Axel-Naumann:
Possibly; I'd like @couet to have a look (for instance at the relative links) once he's back next week.

couet (approved):
It looks good to me. All tests passed. Let see what extra functionalities this will bring.

AdvaitDhingra:
Thanks @couet . Can I squash and merge it?

PR #634, opened 2021-08-28
Configure custom 404 page

ferdnyc:
Someone must've overlooked https://jekyllrb.com/tutorials/custom-404-page/#hosting-on-apache-web-servers — the custom 404 page isn't used on Apache unless it's enabled in the configuration. :wink: 

(Arguably this would be better done in whatever `/etc/httpd/conf.d/` file defines the `VirtualHost` for the site, but those configs aren't part of the repo and it "should" work to define it here as well.)

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,3 +3,4 @@
 	RewriteCond %{REQUEST_URI} /+[^\.]+$ 
 	RewriteRule ^(.+[^/])$ %{REQUEST_URI}/ [R=301,L]
 </IfModule>
+ErrorDocument 404 /404.html


couet:
Thanks for that.

eguiraud:
@couet I think the 404 page worked before but now I get the following:
image

couet:
I’ll check

couet:
I see that on https://root.cern but not on my local build. I will undo this PR to verify if the fact it is not working on https://root.cern comes from this.

eguiraud:
Still broken :/ but it definitely worked at some point, right @Axel-Naumann ?

couet:
Yes it used to work ... until recently

Axel-Naumann:
Fixed (again)!

couet:
Good. should we put back the mod suggested in this PR ?

Axel-Naumann:
We want that 404 page to be global, no need for this PR.

PR #642, opened 2021-09-07
Clarify that available C++ std depends on branch.

Axel-Naumann:
Fixes #8939


eguiraud:
Should we explicitly call out that C++20 is not supported yet?

Axel-Naumann:
I am afraid I'll forget to update that once we support it :-) I think it's reasonable as is. (And 20 will only be supported in master first, etc etc)

PR #649, opened 2021-09-13
Teach {%include ref } to handle nested namespaces:

Axel-Naumann:
{% include ref class="N1::N2::Klass" %} will create <a href=...>N1::N2::Klass</a>, while
{% include ref class="Klass" namespace="N1::N2" %} will create <a href=...>Klass</a>.
Both will link to the documentation of N1::N2::Klass.


couet (approved):
LGTM

PR #650, opened 2021-09-13
Clarify query of ROOT collections.

Axel-Naumann:


Axel-Naumann:
@pcanal FYI

pcanal (commented):
The only thing I would have kept is the mention of TClonesArray which still has a performance edge (less allocations when reading via a TTree) over std::vector.

Axel-Naumann:
Acknowledged but we see little adoption of it (despite its appearance in the Basic section of the manual). It's probably time to say goodbye to advocating for new use, and instead streamline the element allocations for `vector<T*>`, at least in RNTuple?

PR #655, opened 2021-09-13
Page to keep some useful info

couet:
Will be removed once the info there will be moved (or not) elsewhere.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -80,6 +80,8 @@ manual:
         url: manual/geometry
       - title: "Python interface: PyROOT"
         url: manual/python
+      - title: "Developers notes"
+        url: manual/developers_notes
   - title: Further information
     children:
       - title: "Tutorials"
--- a/manual/developers_notes/index.md
+++ b/manual/developers_notes/index.md
@@ -1,0 +1,167 @@
+---
+title: Developers notes
+layout: single
+sidebar:
+  nav: "manual"
+toc: true
+toc_sticky: true
+---
+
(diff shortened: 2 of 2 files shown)


eguiraud:
The "CMake and dictionary" section has been integrated in https://github.com/root-project/web/pull/727

eguiraud:
To be put in a "The ROOT interpreter" page somewhere:

```
### Using a class in the interpreter (cling or PyROOT)

To load a C++ class definition into the interpreter,  it is sufficient to `#include` the header of the class - and probably load the appropriate library.

{% highlight C++ %}
root [0] .L libMyClass
root [1] #include "MyClass.h"
root [2] MyClass obj; obj.DoSomething();
{% endhighlight %}

{% highlight Python %}
import ROOT
ROOT.gSystem.Load('libMyClass')
ROOT.gInterpreter.Declare('#include "MyClass.h"')
obj = ROOT.MyClass()
obj.DoSomething()
{% endhighlight %}

This loads `libMyClass.so` on Linux or macOS, while  `libMyClass.dll` is loaded on Windows.
On macOS, also `libMyClass.dylib` is tried.

You can simplify both the C++ and the PyROOT version by adding the following line to the  `MyClass.h` file.

{% highlight C++ %}
   R__LOAD_LIBRARY(libMyClass)
{% endhighlight %}

This causes the library to be loaded automatically and allows you to skip the `.L` / `ROOT.gSystem.Load()` line.
```

couet:
Replaced by this issue: https://github.com/root-project/web/issues/750

PR #656, opened 2021-09-14
[Docs] Rework section on macros and shared libraries

guitargeek:
* rewrite first part about ROOT macros and ACLiC

  * remove dictionary/cling part to be added back later in functional
    I/O manual

### DIFF
--- a/manual/root_macros_and_shared_libraries/index.md
+++ b/manual/root_macros_and_shared_libraries/index.md
@@ -73,35 +73,33 @@ $ root 'MacroName.C("some String", 12)'
 
 In addition, you can execute a ROOT macro from a ROOT macro.
 
- - [execute a ROOT macro from the invocation of ROOT](#executing-a-root-macro-from-the-invocation-of-root)
-
 <p><a name="executing-a-ROOT-macro-from-a ROOT-macro"></a></p>
 **Executing a ROOT macro from a ROOT macro**
 
-You can execute a ROOT macro conditionally inside another ROOT macro.
+You can execute a ROOT macro conditionally inside another ROOT macro by calling directly the interpreter using
+[TROOT::ProcessLine()](https://root.cern/doc/master/classTROOT.html#a32fc66033a13d1415e0ad523994dd0e5){:target="_blank"}.
 
-- Call the interpreter [TROOT::ProcessLine()](https://root.cern/doc/master/classTROOT.html#a32fc66033a13d1415e0ad523994dd0e5){:target="_blank"}.
-
-`ProcessLine()` takes a parameter, which is a pointer to an `int` or to a
+`ProcessLine()` takes in addition to the code to be executed an optional parameter, which is a pointer to an `int` or to a
 `TInterpreter::EErrorCode` to let you access the interpreter error code after an attempt to interpret.
-This contains the error as defined in enum `TInterpreter::EErrorCode` with `TInterpreter::kSuccess`
-as being the value for a successful execution.
+It returns the return value of the called macro casted to a `Longptr_t`.
 
 _**Example**_
 
-The example {% include tutorial name="cernstaff" %} calls a ROOT macro to build a ROOT file, if it does not exist.
+The example {% include tutorial name="cernstaff" %} calls another macro `cernbuild.C` to build a ROOT file, if it does not exist.
+The function in the `cernbuild.C` macro returns an error code that we get as the return value from `ProcessLine()`.
 
 {% highlight C++ %}
    void cernstaff() {
       if (gSystem->AccessPathName("cernstaff.root")) {
-      gROOT->ProcessLine(".x cernbuild.C");
+         int errorCode = gROOT->ProcessLine(".x cernbuild.C");
+      }
    }
 {% endhighlight %}
 
 
 ## Compiling ROOT macros with ACLiC
 
-ROOT macros are by default just-in-time compiled with Cling based on the Clang compiler. Alternatively, you can use ACLiC to compile your macro from within a ROOT session to a shard library using the system compiler such as gcc.
+ROOT macros are by default just-in-time compiled with Cling based on the Clang compiler. Alternatively, you can use ACLiC to compile your macro from within a ROOT session to a shared library using the system compiler such as gcc.
 
 ACLiC is implemented in [TSystem::CompileMacro()](https://root.cern/doc/master/classTSystem.html#ac557d8f24d067a9b89d2b8fb261d7e18). When using ACLiC, ROOT checks what library really needs to be build and calls your system's C++ compiler, linker and dictionary generator.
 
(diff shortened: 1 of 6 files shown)


manual/root_macros_and_shared_libraries/index.md
```diff
@@ -40,41 +40,39 @@ The name of the ROOT macro and the file name (without file extension) in which t
...
+$ root 'MacroName.C("some String", 12)'
+```
+
+In addition, you can execute a ROOT macro from a ROOT macro.
 
-In addition, you can:
- - [execute a ROOT macro from a ROOT macro](#executing-a-ROOT-macro-from-a ROOT-macro)
  - [execute a ROOT macro from the invocation of ROOT](#executing-a-root-macro-from-the-invocation-of-root)
 
 <p><a name="executing-a-ROOT-macro-from-a ROOT-macro"></a></p>
```
lmoneta, line 78: I would change here: 

**Executing a ROOT macro from a ROOT macro**

You can execute a ROOT macro conditionally inside another ROOT macro by calling directly the interpreter using 
[TROOT::ProcessLine()](https://root.cern/doc/master/classTROOT.html#a32fc66033a13d1415e0ad523994dd0e5){:target="_blank"}.

`ProcessLine()` takes in addition to the code to be executed an optional parameter, which is a pointer to an `int` or to a
`TInterpreter::EErrorCode` to let you access the interpreter error code after an attempt to interpret.
It returns the result of the command casted to a `LongPtr_t`


Then in the example , would it not be better to add a case where we return something ?

manual/root_macros_and_shared_libraries/index.md
```diff
@@ -100,33 +98,21 @@ The example {% include tutorial name="cernstaff" %} calls a ROOT macro to build
...
 {% endhighlight %}
 
-<p><a name="executing-a-root-macro-from-the-invocation-of-root"></a></p>
-**Executing a ROOT macro from the invocation of ROOT**
 
-You can pass a macro to ROOT in its invocation.
+## Compiling ROOT macros with ACLiC
 
-_**Example**_
+ROOT macros are by default just-in-time compiled with Cling based on the Clang compiler. Alternatively, you can use ACLiC to compile your macro from within a ROOT session to a shard library using the system compiler such as gcc.
```
lmoneta, line 104: a ROOT session to a shared library

manual/root_macros_and_shared_libraries/index.md
```diff
@@ -100,33 +98,21 @@ The example {% include tutorial name="cernstaff" %} calls a ROOT macro to build
...
 
-1. Calls `rootcling` to create automatically a dictionary.
+1. Calling `rootcling` to create automatically a dictionary.
 <br/>For creating a dictionary manually, → see [Using rootcling to generate dictionaries manually](#using-rootcling-to-generate-dictionaries-manually).
 
-2. Calls the the system's C++ compiler to build the shared library.
+2. Calling the system's C++ compiler to build the shared library.
 
-3. If there are errors, it calls the C++ compiler to build a dummy executable to clearly report the unresolved symbols.
+3. Load the shared library and optionally execute the macro.
```
lmoneta, line 116: I would remove this following paragraph. I don't think it is true.

manual/root_macros_and_shared_libraries/index.md
```diff
@@ -136,14 +122,13 @@ library by using `#include` statements; you will need to compile each ROOT macro
...
+the classes used in the ROOT macro. Only then you can build and load a shared library containing your ROOT macro.
 
 You can compile a ROOT macro with:
 
   - default optimizations
 
-  - optimizations
+  - different optimizations
 
   - debug symbols
```
lmoneta, line 134: One comment about the note. 

You can compile a ROOT macro that has the function main() but when you execute you cannot call the main() function. 
We should maybe re-phrase this like: 

When a ROOT macro has a function called `main()` your can compile the macro with ACLiC or the Cling ROOT interpreter, but you cannot execute the `main()` function from within the ROOT session.

lmoneta: As a side note I find this a useful functionality Because you can execute at the same time a macro from ROOT and outside ROOT compiling with `root-config`  by doing:
```
void MyMacro() {
.........
} 
int main()  {
  MyMacro();
}
```

lmoneta: I would also move this note at the end of the sub-paragraph after describing the different way of compiling with ACLic a shared library

manual/root_macros_and_shared_libraries/index.md
```diff
@@ -199,385 +180,35 @@ The `$ROOTSYS/include` directory is automatically appended to the include path.
...
 ```
 
 To overwrite an existing include path, type:
 
 ```
-   gSystem->SetIncludePath(" -I$HOME/mypackage/include")
+gSystem->SetIncludePath(" -I$HOME/mypackage/include")
 ```
 
 To add a static library that should be used during linking, type:
```
lmoneta, line 204: I think the example below is for a shared library (not static)

manual/root_macros_and_shared_libraries/index.md
```diff
@@ -199,385 +180,35 @@ The `$ROOTSYS/include` directory is automatically appended to the include path.
...
 
 ```
-   gSystem->AddLinkedLibs("-L/my/path -l*anylib*");
+gSystem->AddLinkedLibs("-L/my/path -l*anylib*");
 ```
 
 For adding a shared library, you can load it before you compile the ROOT macros, by
```
lmoneta, line 210: You can also add a shared library by loading it before compiling the ROOT macros, by doing:

manual/root_macros_and_shared_libraries/index.md
guitargeek: So you mean this paragraph:

```
ACLiC adds the classes and functions declared in included files with the same name as the     
ROOT macro files with one of following extensions: `.h`, `.hh`, `.hpp`, `.hxx`,` .hPP`, `.hXX`.
This means that, by default, you cannot combine ROOT macros from different files into one     
library by using `#include` statements; you will need to compile each ROOT macro separately.
```

Right let's remove it, I also think it's too specific to the basic manual.

manual/root_macros_and_shared_libraries/index.md
guitargeek: Good to know!

manual/root_macros_and_shared_libraries/index.md
guitargeek: Ok, I fixed this typo.

manual/root_macros_and_shared_libraries/index.md
guitargeek: Okay I'll change it

manual/root_macros_and_shared_libraries/index.md
guitargeek: Right, sounds better

manual/root_macros_and_shared_libraries/index.md
guitargeek: Ok good point, I have changed the example to pretend that `cernbuild` returns some error code that we will store in a new variable.

PR #673, opened 2021-09-16
[Docs] Remove outdated section about portable ROOT macros

guitargeek:
These instructions applied to cint in the past, but not to cling.

Also, some more notes are added to the `root_macros_and_shared_libraries` section.


lmoneta (approved):
Looks good .Thanks Jonas

PR #713, opened 2022-01-04
Fix heading for "Concluding Remarks"

simonthor:
See the issue here: https://github.com/root-project/root/issues/9493, which I mistakenly opened in the wrong repo.


eguiraud:
Thank you very much @simonthor , will merge if green!

eguiraud (approved):

PR #722, opened 2022-01-17
Manual: physics vectors

Axel-Naumann:


manual/physics_vectors/index.md
```diff
@@ -7,762 +7,50 @@ toc: true
 toc_sticky: true
 ---
 
-ROOT provides several packes for physics vectors:
+Manual: physics vectors
+=======================
```
hahnjo, line 11: I think the heading is automatically added by the infrastructure, I only added it for the CodiMD

hahnjo (approved):
otherwise LG

PR #739, opened 2022-01-21
Rework TMVA manual section with a broader focus

vepadulano:
This PR adds a small introduction with sections about machine learning with ROOT. Also focuses on showing interoperability with other tools (loading external models, exporting ROOT data to Numpy).
This is just a first step. As discussed with Lorenzo, more sections (especially one about SOFIE) can be added later if this part looks ok

### DIFF
--- a/manual/tmva/index.md
+++ b/manual/tmva/index.md
@@ -1,5 +1,5 @@
 ---
-title: TMVA
+title: Machine learning with ROOT
 layout: single
 sidebar:
   nav: "manual"
@@ -7,10 +7,15 @@ toc: true
 toc_sticky: true
 ---
 
-The Toolkit for Multivariate Data Analysis with ROOT (TMVA) provides a machine learning
-environment for the processing and evaluation of multivariate classification, both binary
-and multi-class, and regression techniques targeting applications in high-energy physics.
-The package includes:
+Machine learning plays an important role in a variety of HEP usecases. ROOT
+offers native support for supervised learning techniques, such as multivariate
+classification (both binary and multi class) and regression. It also allows easy
+interoperability with commonly used machine learning libraries.
+
+## TMVA
+
+TMVA is the ROOT library that provides the interfaces and implementations of the
+above mentioned machine learning techniques. The package includes:
 
 - Neural networks
 - Deep networks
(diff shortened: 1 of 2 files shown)


manual/tmva/index.md
```diff
@@ -1,16 +1,21 @@
...
-The package includes:
+Machine learning plays an important role in a variety of HEP usecases. ROOT
+offers native support for supervised learning techniques, such as multivariate
+classification (both binary and multi class) and regression. It also allows easy
+interoperability with commonly used machine learning libraries.
+
+## TMVA
+
+TMVA is the ROOT library that provides the interfaces and implementations of the
+above mentioned machine learning techniques. The package includes:
```
lmoneta, line 18: The list below needs to be updated.. 
I can do this in a later PR

manual/tmva/index.md
```diff
@@ -26,29 +31,77 @@ The package includes:
...
+   // load trees into TMVA
+   TMVA::DataLoader loader{"mydataloader"};
+   loader.AddSignalTree(signaltree);
+   loader.AddBackgroundTree(backgroundtree);
+}
+{% endhighlight %}
+
+The loaded data can then be passed to the {% include ref class="Factory"
+namespace="TMVA" %} class for training, as shown in
+[this tutorial](https://root.cern/doc/master/tmva003__RReader_8C.html){:target="_blank"}.
```
lmoneta, line 59: I think this is not the right tutorial. The Reader is for inference. 
I think the tutorial should be `TMVAClassification.C`

manual/tmva/index.md
```diff
@@ -26,29 +31,77 @@ The package includes:
...
+## Interoperability with machine learning libraries
+
+Machine learning is a widely researched topic. Many libraries implement
+classification and regression techniques, with a broader scope or in more
+specific fields. These libraries also vary in the programming languages offered
+by their APIs, although it is true that often Python defines a common ground.
+Notable examples include [Keras](https://keras.io/){:target="_blank"},
+[PyTorch](https://pytorch.org/){:target="_blank"} and
+[scikit-learn](https://scikit-learn.org){:target="_blank"}.
+
```
lmoneta, line 70: I would add here a paragraph: 

#Training in TMVA using an external library

Here is good to link the ClassificationKeras tutorial

manual/tmva/index.md
```diff
@@ -26,29 +31,77 @@ The package includes:
...
+### Loading an external model into TMVA
+
+Suppose you have trained a model using one of the external libraries cited above.
+Or maybe you need to use a specific pre-trained model for your usecase. TMVA
+allows loading model files directly in a factory object and use it for example
+to classify a test dataset. In particular, the entire application could be
+written as a single Python script thanks to
+[PyROOT]({{ '/manual/python' | relative_url }}). A good example of this can be
+found in [this tutorial](https://root.cern/doc/master/ClassificationKeras_8py.html){:target="_blank"}.
+
```
lmoneta, line 80: The example for using a pre-trained model on a test data set should be: 
- https://root.cern.ch/doc/master/ApplicationClassificationKeras_8py.html  (when using old TMVA::Reader interface)
- https://root.cern.ch/doc/master/tmva003__RReader_8C.htmlhttps://root.cern.ch/doc/master/tmva003__RReader_8C.html
  using the new Reader interface.
  
  Probably it would be nice to link only the RReader tutorial, but we need to translate it in Python (and implement a direct conversion numpy->RTensor)

lmoneta (approved):
Thank you Vincenzo for writing this introduction.
I think it is a good start to improve then the TMVA doc.
I added some comments, but I can include myself in a following PR

manual/tmva/index.md
vepadulano: Sure, it's the section below titled "Loading an external model into TMVA". I can change the section title if you want, let me know

manual/tmva/index.md
vepadulano: Now that I read the other tutorials I get better the source of confusion. in the https://root.cern/doc/master/ClassificationKeras_8py.html the example is just creating a model with `Keras`, then doing the training in TMVA. Your examples are the right ones for pre-trained models instead. I will change this section accordingly, showing only the `ClassificationKeras` example. For the other examples, we can do them in another PR

manual/tmva/index.md
lmoneta: Sounds good!

manual/tmva/index.md
lmoneta: Agree, we can add example for inference, including also SOFIE, in another PR

lmoneta:
I think we can merge it now

PR #756, opened 2022-03-02
make QtCreator post easier to follow

ferdymercury:
Fill gaps in the explanation found by couet.


couet (approved):

couet:
LGTM ! Thanks for that.

PR #758, opened 2022-03-03
remove broken link

ferdymercury:
6.24 does not have qch file


couet (approved):

couet:
There is an error: https://github.com/root-project/web/runs/5406161326?check_suite_focus=true

ferdymercury:
That error seems to come from a page I didn't modify: ./build/758/get_started/courses/index.html

couet:
It is weird because the link to "columbia" is valid... I re-launched it to see if it was temporary unavailable

couet:
Ok it passed

PR #766, opened 2022-03-07
executing a multi-line command did not reflect the reality

couet:


ferdymercury:
I think the numbers between brackets don't need to be added, because earlier in the manual, it is said:

`For the further examples we will "abbreviate" root [0] etc by root [].`

ferdymercury (changes requested):
Line 140 should be removed, too, I guess?

couet:
oops sorry I metered ...

ferdymercury:
It's really weird why this was there in the first place. I just tried with ROOT 5.34 and CINT, and the question marks don't work there, either... :

```
? TLine l;
Syntax Error: ? TLine l (tmpfile):1:
*** Interpreter error recovered ***

ROOT special commands.
===========================================================================
             pwd          : show current directory, pad and style
             ls           : list contents of current directory
             which [file] : shows path of macro file


```

PR #767, opened 2022-03-07
cling example was wrong

couet:


manual/cling/index.md
```diff
@@ -127,18 +127,16 @@ filename     line:size busy function type and name
...
 root [] .>
-root [] ?
 {% endhighlight %}
 
 Here we see:
 
 -   Use `.class` as quick help and reference
--   Unix like I/O redirection using `.> out.txt` and unredirection with `.>`
--   Use `?` to get help on all ‘‘raw'' interpreter commands
--   Use @ to abort a multi-line command
```
ferdymercury, line 141: This one is still valid, by changing to `.@`

couet: This PR is correct now. Do you agree ? I checked it . the TLine `l`was not not created either....
Do you agree I merge it ?

ferdymercury: I would restore line 141, but changing @ with .@
.@ works well for interrupting a multi-line command.

couet: I remove it because it is better explained in the next example

ferdymercury: oh I see.

And what about removing the number within the root[ ] brackets?

manual/cling/index.md
```diff
@@ -127,18 +127,16 @@ filename     line:size busy function type and name
 (compiled)     (NA):(NA) 0 public: static int ImplFileLine();
 (compiled)     (NA):(NA) 0 public: static const char *ImplFileName();
 (compiled)     (NA):(NA) 0 public: static int DeclFileLine();
+root [] TLine l;
 root [] .> test.log
-root [] l.Dump();
+ l.Dump();
```
ferdymercury, line 132:
```suggestion
root [] l.Dump();
```

PR #775, opened 2022-03-21
mention full flags and build directory

ferdymercury:


couet (approved):
LGTM. I Think we can merge ?

PR #780, opened 2022-04-20
python2 ipy needed for notebooks

ferdymercury:
as found out by couet

### DIFF
--- a/for_developers/doxygen/index.md
+++ b/for_developers/doxygen/index.md
@@ -31,8 +31,7 @@ browser, the file `${HOME}/rootdoc/html/index.html`.
 
 ### Compilation prerequisites on Ubuntu
 - apt install doxygen make jupyter-nbconvert 
-- pip install scandir nbformat metakernel
-- python2 -m pip install ipykernel
+- pip install scandir nbformat metakernel ipykernel
 - Optional: apt install qhelpgenerator-qt5. In case of older Ubuntu versions, use qt4-dev-tools instead, and modify in the Doxyfile QHG_LOCATION to /usr/lib/x86_64-linux-gnu/qt4/bin/qhelpgenerator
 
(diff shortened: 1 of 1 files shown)


for_developers/doxygen/index.md
```diff
@@ -32,6 +32,7 @@ browser, the file `${HOME}/rootdoc/html/index.html`.
 ### Compilation prerequisites on Ubuntu
 - apt install doxygen make jupyter-nbconvert 
 - pip install scandir nbformat metakernel
+- python2 -m pip install ipykernel
```
eguiraud, line 35: uhm, shouldn't python3 be the default these days?

ferdymercury: Hmmm the Makefile is currently not forcing the user to have a minimum python version. There is now:
`Makefile:PYTHON_EXECUTABLE ?= /usr/bin/python`

The root make-doc build machine has /usr/bin/python pointing to python2, as compared to /usr/bin/python3, so couet found that installing this extra package solved the many errors that were being found in the console log https://lcgapp-services.cern.ch/root-jenkins/job/root-makedoc-master/2651/consoleFull jupyter_client.kernelspec.NoSuchKernel: No such kernel named python2

(That being said, it would be also fine for me to change the Python_executable in the Makefile to python3).

eguiraud: i can't comment as to why the machine where we build the docs uses python 2, but most developers and users these days are on python 3 (and they should be, python 2 is dead since january 2020). so when explaining to new devs how to generate the doxygen docs, we should not impose python 2 -- we can just write `python` here.

ferdymercury: Good point. I merged it with the line above.

couet (approved):
LGTM

PR #785, opened 2022-04-27
Build from source instructions: add `--depth=1` option to `git clone`

jalopezg-r00t:
A shallow clone will benefit users who want to download a snapshot of the latest release, as it will avoid downloading objects that are not part of the tree for the latest commit.

For users that want to stay updated via `git pull`, this will be delayed until the first `git pull`.

See https://github.com/root-project/root/issues/10238 for context.


eguiraud (approved):
good for me!

PR #799, opened 2022-08-05
Update copyright year

etejedor:


petrstepanov:
It is better to use `{{ 'now' | date: "%Y" }}` instead of a static year in HTML. This way the copyright year will always render current year.

couet (approved):
Thanks !

etejedor:
@couet feel free to address @petrstepanov's comment in another PR.

PR #811, opened 2022-10-05
Rewrite intro of QtCreators blog post

ferdymercury:


couet (commented):
I am not sure that "zooming a graph" is a good example of why ROOT is "difficult to learn". You point to an example that shows how to do it in a script. Fine. And I suppose you mean that the other packages are easier to use because you can do it with the mouse? But ROOT allows also you to do it with the mouse. So zooming a graph is really super intuitive and simple with ROOT.

ferdymercury:
Hi Olivier, thanks for the feedback. I was speaking from the perspective of someone coding a script, not from the GUI point of view, which is indeed easy.

Whenever I want to zoom a graph in a script, I need to open this example because it's non-intuitive to remember what was the step needed for this. (To draw a separate frame in advance in the canvas using c.DrawFrame function). Maybe it's more 'hard to remember' rather than 'hard to learn'.

In comparison, in matlotlib you just use `plt.xlim(0,3)`, which is very intuitive and almost impossible to forget, you never need to come back to that example in the future.
Or zooming a histogram in ROOT is also much easier, SetRangeUser(.., ...) than zooming a graph, this one is very intuitive too. It's also quite confusing for students to remember that they need to use different functions when zooming a graph compared to when zooming a histogram. That 'adds' to the learning curve, or the 'remembering curve'.

I speak not just based on my own experience, but based on the different students I have, and the 'historically most asked/repeated questions'.

PR #814, opened 2022-10-07
Force download for gz, pkg, exe and zip

SimonWaldherr:
Force download for gz, pkg, exe and zip files in .htaccess

the download only works for me if I right click on the download links and click "Download Linked File". Otherwise, the browser loads the file and tries to display it:
root cern

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -3,3 +3,8 @@
 	RewriteCond %{REQUEST_URI} /+[^\.]+$ 
 	RewriteRule ^(.+[^/])$ %{REQUEST_URI}/ [R=301,L]
 </IfModule>
+# Force download for gz, pkg, exe and zip
+<FilesMatch "\.(gz|pkg|exe|zip)$">
+   ForceType application/octet-stream
+   Header set Content-Disposition attachment
+</FilesMatch>


couet:
Sorry to not have seen it before. I will check

couet:
I  do not see the effect you are observing. When I try to download something from https://root.cern/releases/release-62610/ it is ok. Don't you think it is not due to some setup on your local web browser?

SimonWaldherr:
Thanks @couet.
Apparently only *.pkg-files on Safari on macOS are affected here. I could not reproduce the problem with other browsers. However, it also occurs on another device of mine with a macOS clean install.
Actually, I would have preferred to force the download with the [HTML Download attribute](https://caniuse.com/download), but I couldn't find where this should be added.

Axel-Naumann:
@SimonWaldherr could you try again? Should be fixed now. (I cannot reproduce it to begin with, so I need to rely on your confirmation - but I was able to see the HTTP header being "better" now than before.)

SimonWaldherr:
Unfortunately, the problem still occurs with macOS Safari. For me, only ```.pkg``` files are affected.
Screenshot 2022-12-12 at 06 56 43 PM

Axel-Naumann:
Super strange. The MIME type is now correct (thanks for verifying that!) and we cannot reproduce this anywhere. I really wonder what's happening here. Do you get the same result with `http://`?

Which macOS version is this?

SimonWaldherr:
I tested it with macOS 11 and macOS 12, unfortunately I can't update to macOS 13 at the moment.
The same behavour with the firewall switched off. It doesn't matter whether I use DSL or an LTE hotspot.

Let's drop it, it doesn't matter if it only occurs on my machine, there is a workaround via "Download Linked File".

Thanks anyway.

PR #830, opened 2022-12-09
new publications page

couet:
new publications mechanism

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -298,6 +298,8 @@ collections:
     output: true
   ideas:
     output: true
+  publications:
+    output: true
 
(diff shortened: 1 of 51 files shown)


eguiraud:
This:

image

needs to say "from root-project:new-publications-page" rather than "from couet:new-publications-page" in order to get the preview. You need to push the branch to the upstream repo and then open a PR for that branch.

couet:
I tried Friday to make it "previewable". It did not work. I will explain/show it during the meeting today. Like that, everybody will see what I propose (not only the 3 reviewers). It will be easier.

PR #841, opened 2023-01-17
Fix description of `target` for I/O customization rules

jalopezg-git:
Apparently, according to [core/meta/src/TSchemaRule.cxx#L528](https://github.com/root-project/root/blob/master/core/meta/src/TSchemaRule.cxx#L528) and the implementation of `TSchemaRule::ProcessList()`[1], the list is **comma-**separated.

[1] https://github.com/root-project/root/blob/master/core/meta/src/TSchemaRule.cxx#L953


pcanal (approved):
Thanks.

PR #853, opened 2023-03-24
prevent ambiguity in what latest release means

ferdymercury:
As reported by several users:
https://root-forum.cern.ch/t/latest-root-release/52243

Make the label more explicit.


couet (approved):
LGTM

PR #865, opened 2023-06-27
Download and use OpenSans fonts directly from root.cern website

linev:


linev:
Use branch in repository itself to check if it works.
See #866

PR #866, opened 2023-06-27
Download and use OpenSans fonts directly from root.cern website

linev:


linev:
I close this PR - changing only fonts does not help much.
There are much more resources download from google or third-party websites.
One need to find other solution

Axel-Naumann:
>  There are much more resources download from google or third-party websites.

Can we do one by one? Isn't this already an improvement compared to before?

Axel-Naumann:
@dpiparo FYI!

PR #868, opened 2023-07-05
Remove mention of the conda nightlies

eguiraud:
Installing the conda nightly package actually gets the user a build from November '22, because the Jenkins nightly builds have been broken since then.

I'm not sure there is progress in fixing the conda nightly builds soon, so I would propose, in the meanwhile, to not suggest this as a method to get a ROOT nightly build.

RFC:
- @chrisburr 
- @vepadulano


vepadulano (approved):
Let's hope we can revert this change soon :D

PR #872, opened 2023-07-25
Fix #13276 by fixing the typo on the x axis of a plot

dpiparo:


ferdymercury:
Thanks Danilo!! This file would also need replacement: https://github.com/root-project/web/blob/main/primer/graph_with_expectation.png

ferdymercury:
@couet thanks!! what about https://github.com/root-project/web/blob/main/primer/graph_with_expectation.png

PR #879, opened 2023-09-12
Remove service worker

dpiparo:


dpiparo:
THe PR must be made from this very repo and not from a fork in order to trigger the test generation of the site: see https://github.com/root-project/web/pull/880

PR #884, opened 2023-09-13
Fix mistake in publication title

dpiparo:
because the yaml cannot be parsed as is by jekyll.


couet (approved):
Ah yes the `:` in the title will be assumed as a new field. I missed that one. Thanks ! :-)

PR #886, opened 2023-09-13
Update status of ex-team member

dpiparo:


_data/authors.yml
```diff
@@ -299,9 +299,13 @@ Danilo Piparo:
                  to many of its areas: I/O, reflection, math and statistics. He was mainly
                  involved in the aspects of ROOT related to parallelism,
                  performance and programming model."
-  tag         : "previous"
+  short_bio   : "Axel is leading the ROOT project since 2017."
```
pcanal, line 302: Is that line in the right place?

dpiparo: no indeed a bug.

PR #888, opened 2023-09-13
swap binary and source distributions

ferdymercury:
many newbies go to the Download ROOT instructions and the first thing they do is download the source code and try building it, because it appears at the very top of the page. Swap the order to make it less likely to happen.

### DIFF
--- a/_releases/make-release-page.py
+++ b/_releases/make-release-page.py
@@ -302,6 +302,16 @@ def generateDownloadInfo(outfile, version, downloadInfo):
     """
     Write out source and binary download links.
     """
+    outfile.write(F"""
+## Binary distributions
+
+| Platform       | Files | Size |
+|-----------|-------|-----|
+""")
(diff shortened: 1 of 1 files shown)


_releases/make-release-page.py
```diff
@@ -302,6 +302,16 @@ def generateDownloadInfo(outfile, version, downloadInfo):
     """
     Write out source and binary download links.
     """
+    outfile.write(F"""
+## Binary distributions
+
+| Platform       | Files | Size |
```
Axel-Naumann, line 308: It would be even better to call out that the package manager might already have it: apt/dnf/brew/conda install root-system or cerm-root might do. (We should lost the actual commands for these four package managers.)

Do you think you could provide this in a follow-up PR? This one is very nice as is!

Axel-Naumann (approved):
Thanks!

PR #897, opened 2023-09-26
Prioritize some installation way for most users

ferdymercury:


ferdymercury:
Hi Axel, I added some notes on package managers as you suggested in the previous PR. Cheers!

PR #976, opened 2024-03-25
Improve manual on graphics with python

jblomer:
Point out that graphics of function-local objects disappears when the function ends.


jblomer:
@couet For this repository, do we wait for review or are we merging at our own discretion?

dpiparo (approved):

PR #977, opened 2024-03-30
Update ClangFormat

pitkajuh:
Update the ClangFormat to match the one in main Root repository. Please, see https://github.com/root-project/root/blob/master/.clang-format


guitargeek:
Thanks for the initiative, nice catch!

However, I think it's more sustainable to just link the actual `.clang-format` file on the website, which I suggest in this PR:
https://github.com/root-project/web/pull/984

PR #981, opened 2024-04-05
document run from shared object

ferdymercury:
matches https://github.com/root-project/root/pull/14813/


dpiparo (approved):
LGTM, thanks for the changes!

PR #997, opened 2024-05-28
Mention Powershell in the documentation

ferdymercury:
thisroot.ps1 was added long ago: https://github.com/root-project/root/issues/6815

but the documentation on the website was not updated, so it's hard to know that it exists, and multiple Window users in the forum have asked why running ROOT from within Visual Studio terminal does not work.


bellenot (approved):
Thanks Fernando!

PR #999, opened 2024-05-28
Update make-release-page.py

ferdymercury:
https://root.cern/install/all_releases/
and
https://root.cern/releases/release-63200/

still show the slash instead of the dot


ferdymercury:
pinging @dpiparo

dpiparo:
Apologies for noticing this only now. Excellent proposal.

PR #1010, opened 2024-07-05
Change the recommended way of getting a TTree from a TFile in python

jmcarcell:
After https://github.com/root-project/root/pull/14660 the current way is deprecated.


guitargeek:
Great thanks!

PR #1017, opened 2024-08-27
Update dependencies.md

ferdymercury:
see https://github.com/root-project/root-docker/blob/master/ubuntu2404/packages


dpiparo:
Thanks!

PR #1020, opened 2024-08-28
Blog post about the student course

martamaja10:


dpiparo (approved):
Thanks, the post looks really nice. I have a doubt: all other posts have a longer "preview" which is shown on the main page of the website. Would it make sense to extend it a bit also for this post?

martamaja10:
> I have a doubt: all other posts have a longer "preview" which is shown on the main page of the website. Would it make sense to extend it a bit also for this post?

Automatically only the first paragraph is used as <post.excerpt>, in the new commit I added the way to modify it per blog post in case we want to copy it for the next blog posts. Now two short paragraphs will be included in the "preview" on the main page.

PR #1035, opened 2024-11-12
Correctly parse the new Windows binaries names (not a change of the website but of the script that builds release pages)

dpiparo:
Before this improvement https://github.com/root-project/root/commit/1998f735fe7094e1f7fc5078f50ca40030426262 they had names like "root_v6.35.01.win64.vc17.zip". After this improvement, i.e. 2024-11-7, "root_v6.35.01.win64.python311.vc17.zip" i.e. they encode the python version.

### DIFF
--- a/_releases/make-release-page.py
+++ b/_releases/make-release-page.py
@@ -261,12 +261,12 @@ def binaryNameToTitle(filename):
 
         return F"macOS {macOSVers} {arch} Xcode {xcodeVers}"
     elif platformCompiler.startswith('win32') or platformCompiler.startswith('win64'):
-        regexWin = re.compile(r'^win(32|64)[.](vc[0-9]+)(.debug)?')
+        regexWin = re.compile(r'^win(32|64)[.](python3[0-9]+)[.](vc[0-9]+)(.debug)?')
         matchWin = regexWin.match(platformCompiler)
         if not matchWin:
             print(F'ERROR: cannot parse Win32 MSVC version for {filename}')
             return filename
-        (bitness, vcVers, debug) = matchWin.groups()
+        (bitness, pythonVers, vcVers, debug) = matchWin.groups()
         if not debug:
             debug = ''
         else:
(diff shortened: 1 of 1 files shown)


_releases/make-release-page.py
```diff
@@ -261,12 +261,12 @@ def binaryNameToTitle(filename):
 
         return F"macOS {macOSVers} {arch} Xcode {xcodeVers}"
     elif platformCompiler.startswith('win32') or platformCompiler.startswith('win64'):
-        regexWin = re.compile(r'^win(32|64)[.](vc[0-9]+)(.debug)?')
+        regexWin = re.compile(r'^win(32|64)[.](python3[0-9]+)[.](vc[0-9]+)(.debug)?')
```
bellenot, line 264: We don't build `debug` anymore. Should we distribute `relwithdebinfo`? Or do we distribute only `release` binaries?

_releases/make-release-page.py
```diff
@@ -291,7 +291,7 @@ def binaryNameToTitle(filename):
             bitArch = "x86"
         else:
             bitArch = "x64"
-        return F"Windows Visual Studio {versYear} {bitness}-bit {bitArch} {debug}"
+        return F"Windows Visual Studio {versYear} {bitness}-bit {bitArch}{debug}"
```
bellenot, line 294: Same comment here

bellenot (changes requested):

_releases/make-release-page.py
dpiparo: So far we have distributed release binaries: this can be changed if we believe that relwithdebinfo is useful to users. The regex is at the moment harmless, in the sense that in case debug binaries are available, the markdown of the page is generated accordingly. To be 100% correct, one should indeed detect the last part of the binary name and attach it to the name accordingly: release by default, debug in case we have debug binaries and release with debug info (see above necessity criterion)

PR #1049, opened 2024-11-25
expand example using `@pythonize`

m-fila:
Expand example using `@pythonization`. The older example was very minimal and not even showing a class that is getting pythonized.

Part of the https://github.com/orgs/root-project/projects/18?pane=issue&itemId=88465275


vepadulano (approved):
Thank you!

PR #1099, opened 2025-04-03
Update opensuse dependencies

ferdymercury:
see https://root-forum.cern.ch/t/help-building-root-from-source-opensuse-tumbleweed/57141/7?u=ferhue


dpiparo:
Thanks!

PR #1116, opened 2025-06-01
Write blog post about building ROOT on macOS with Nix

guitargeek:
Showing a streamlined platform-independent way to do ROOT development could help to attract more development talent.


dpiparo (approved):
Thanks for this initiative!

PR #1118, opened 2025-06-02
Mention ClassDefOverride

ferdymercury:


manual/io_custom_classes/index.md
```diff
@@ -217,6 +217,8 @@ The syntax is:
 ClassDef(<ClassName>, <VersionNumber>)
 {% endhighlight %}
 
+(In the case of classes inheriting from `TObject`, it's better to use `ClassDefOverride` instead of `ClassDef`.)
```
pcanal, line 220: More than inheritance from `TObject` the choice between `ClassDef` and `ClassDefOverride` is based on whether the developer of the class has already transitioned to using the `override` keyword.  (If I remember correctly using `ClassDefOverride` for classes that does not use the override keyword will result in warning and vice et versa).

ferdymercury: Mmm are you sure? This is just for classes deriving from TObject derived functions (CheckConsistencyHash and so on)

pcanal: You are right it is more complicated.  It need both the current class using the keyword *and* (one of) the base class(es) having a `ClassDef` (Otherwise, indeed, it is the first declaration of those function and it needs to use the `virtual` keyword).

ferdymercury:
```suggestion
(In the case of classes inheriting from `TObject`, or from a non-TObject base class declaring itself `ClassDef*`,
it's then better to use `ClassDefOverride` in the child instead of `ClassDef`.)
```
fyi @guitargeek

ferdymercury:
Yes please :)

silverweed:
@ferdymercury I think the CI is failing due to missing credentials. This is expected if you opened this PR from a fork (see [here](https://github.com/root-project/web?tab=readme-ov-file#test-deployment-of-pull-requests)): if that's the case, could you please reopen it by branching the main repo instead? I'll approve it and we can merge it once the CI passes

silverweed:
Ah, I see, then I guess I'll do it. 
Thanks for the contribution!

PR #1136, opened 2025-07-07
remove obsolete section about code formatting, explain clang-format better

silverweed:


contribute/coding_conventions/index.md
```diff
@@ -98,123 +98,24 @@ Note the mandatory method separator line:
 {% endhighlight %}
 exactly 80 characters long.
 
-## Preferred Coding Style
-Here we describe our preferred coding style. Coding style is very personal and we don't want to force our views on anybody. But for any contributions to the ROOT system that we have to maintain we would like you to follow our coding style.
-
-### Indentation
-To be able to keep as much code as possible in the visible part of the editor of to avoid over abundant line wrapping we use indentation of 3 spaces. No tabs since they give the code always a different look depending on the tab settings of the original coder. If everything looks nicely lined up with a tab setting of 4 spaces, it does not look so nicely anymore when the tab setting is changed to 3, 5, etc. spaces.
```
pcanal, line 105: Consider keeping those 2 non-tool specific paragraphs.

silverweed: The first may still be useful indeed, but the second is completely handled by the formatter, no?

pcanal: True but clang-format is not yet (there is hope though) fully supporting the 'real' formatting we want (see in particular the alignment in the header files), consequently we are still needs to leave the clang-formatting optional and thus some unwanted formatting will still be committed.  Most formatting are harmeless-ish and can be most hand done ... the tab vs space is extremely (slight exaggeration) annoying (to me) and can be invisible to (newer) developers, so I would keep it.

On a different note, there is a slight question on whether it is helpful to describe the general/main formatting choices so developers get used to typing it almost correct ... or not. (but we definitively can not expect developers to read the clang-format to know the rules (the .clang-format is too dense and long) ... on the other hand the existing might be a (close enough?) documentation/example)

silverweed: I agree that it is good to give developers an overview of the main formatting choices and their rationale; on the other hand the use of an automated formatter should have the main goal of relieving them from having to spend time thinking about it. What about keeping the paragraphs but placing them below the clang format description? And make it clear that if all they care about is submitting "valid" code they can skip them but if they want more details they can optionally read them...

pcanal: Yes, reordering in this way sounds fine.

silverweed:
@pcanal Do you think this can be merged after the latest changes?

silverweed:
Sorry, once again opened from the wrong repo. New PR: https://github.com/root-project/web/pull/1149

PR #1143, opened 2025-08-19
Add build options for ROOT 6.38 and improve NixOS install instructions

guitargeek:


vepadulano:
The build options table should be generated via the script during the release process

PR #1146, opened 2025-08-28
Add selectversions.js to version control and update it

silverweed:
See the commits for more details.

**TL;DR**: this allows us to support `"latest-stable"` as a pseudo-version in the doc but requires some little manual intervention during the release procedure. The steps to add to the release procedure are the following:
1. in `selectversion.js`, change `LATEST_VERSION` to the new release (e.g. `638`);
2. on eos, change the `latest-stable` symlink to point to the new latest release.

These modifications are already live and you can see them here: [https://root.cern/doc/latest-stable/](https://root.cern/doc/latest-stable)


ferdymercury:
> em here: [https://root.cern/doc/latest-stable/](https://root.cern/doc/latest-stable)

Thanks!
Just as a comment, it's not easy to see what latest-stable version corresponds to.

image

It would be great if this box could have also in parenthesis (v634) or whatever.

silverweed:
@ferdymercury That's not supposed to happen: it should redirect you to 636...which browser are you using?

PR #1150, opened 2025-09-01
Reorganize logos, add svgs

silverweed:
- Add logos to the repo
- Add SVG versions of the logos
- Remove external link to Google Drive
- Rename logo files to be more coherent
- Reorganize directory structure of logos. Now all logos are under `assets/images/logos` and are divided into `icons`, `png`, and `svg`.
- Use svg for embedded images in the logos page where it makes sense (i.e. where they are smaller than their png counterpart)

**NOTE**: when this is merged we need to manually remove the `img/logos` folder from eos. There are a couple of images in there that don't seem to belong anywhere (`final-splash-mask-*.gif`) - not sure if we want to keep them...


enirolf (approved):
This is super nice! Maybe not for this PR, but it could potentially also be nice to have a button/link that allows users to download an archive with all logos.

I will let @dpiparo and/or @couet comment on the note you made.

PR #1168, opened 2025-11-18
Update what_does_root_stand_for.md

ferdymercury:
fyi @couet @silverweed


ferdymercury:
Is there a way to exclude the root.cern link rule for these old links?
Sth like rule:off xxx  rule:on comment ?

Issue #1018, opened 2024-08-27
Rework dependencies


ferdymercury:
Feature suggesstion:

The dependencies webpage is hard to maintain. Instead, I suggest the following:

I propose instead to auto-paste in the ROOT webpage the packages listed in the root-docker image packages file, by using js:

https://stackoverflow.com/questions/71484264/can-you-get-the-content-of-a-github-file-with-javascript

PR #1238, opened 2026-08-11
Simplify dependencies by pointing to Docker packages

ferdymercury:
Dependencies are always outdated and  unmaintainable. One is better off at pointing towards Dockerfile, even if sometimes an overkill or not perfect since build and runtime dependencies are not the same. In any case, git is better since at least you see how the dependencies change over time / release versions compared to just the current snapshot of the website.

### DIFF
--- a/install/dependencies.md
+++ b/install/dependencies.md
@@ -31,10 +31,14 @@ we recommend checking the [docker image repository](https://github.com/root-proj
 - GCC 4.8 or later is supported
 - Clang 3.4 or later is supported
 
-## Fedora, Scientific Linux and CentOS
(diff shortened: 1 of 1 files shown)


ferdymercury:
fyi @hageboeck @guitargeek @silverweed

silverweed:
Should we remove the old instructions perhaps?

ferdymercury:
> Should we remove the old instructions perhaps?

Fine for me, but I thought the "details" tag hides them enough? Maybe as a first transition to see how people react to it ?

ferdymercury:
Fixes https://github.com/root-project/web/issues/1018

